PR #148, opened 2014-07-05
fixed #143, unset keybindings will now show striked-through in both keyb...

### LINKED ISSUE #143: Keybinding shows for a package even when unset
For example, I have the following in my `keymap.cson`:

``` cson
'.autocomplete-plus input.hidden-input':
  'tab': 'unset!'
  'enter': 'autocomplete-plus:confirm'
```

But this shows in the Settings View for Autocomplete Plus:

screenshot_2014-06-18_12_32_46

Perhaps just rendering the entry with strikethrough would be best because you could see that it is there by default, but isn't enabled right now.

Azeirah:
...indings panel and in package-keybindings

### DIFF
--- a/lib/keybindings-panel.coffee
+++ b/lib/keybindings-panel.coffee
@@ -86,14 +86,23 @@ class KeybindingsPanel extends View
     {selector, keystrokes, command, source} = keyBinding
     source = @determineSource(source)
     $$$ ->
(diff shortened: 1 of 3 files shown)


kevinsawicki:
Closing this out until it is updated to address the comments, please comment when it is ready for another look.

PR #272, opened 2014-10-27
navigate with up & down

dnozay:


mnquintana:
@kevinsawicki Does this need anything else to get merged?

mnquintana:
@dnozay Would you mind adding specs for this?

50Wliu:
Also @dnozay the Atom team tries to avoid `&&`, `||`, `==`, as much as possible and replaces them with `and`, `or`, etc.  Changing it here would break the formatting guidelines in the rest of Settings View, which is probably why @philipgiuliani is suggesting for you to move away from those.

dnozay:
@mnquintana, don't know how to simulate the `core:move-up` / `core:move-down` events.
probably need tests and fixtures for

* no settings to cycle through
* mix of visible / hidden items in the panel - cycle should work properly
* first / last item active and try to cycle around.

I would prefer for someone who's more familiar with the tests to add some.

mnquintana:
> don't know how to simulate the core:move-up / core:move-down events.

For that you can use [`atom.commands.dispatch settingsView.element, 'core:move-up'`](https://atom.io/docs/api/v0.196.0/CommandRegistry#instance-dispatch) (where `settingsView.element' is the element of Settings View, defined somewhere).

No problem though if you're not comfortable writing tests right now - I can take a look at it for now.

mnquintana:
@dnozay Actually, this would need some more changes to get merged. Currently, the keyboard shortcuts only work if you manually focus on the Settings View after it loads (ie. by clicking on an element inside of it), and switching between some panels can make the focus change so the keybindings no longer work.

benogle:
I think the up and down keys should scroll the settings view like a web page. We should have an alternative command for cycling through the panels. Maybe `settings-view:next-panel` and `settings-view:previous-panel`. I like the slack keybinding for this: `option-up` and `option-down`.

So my vote is :-1: on core:move-up / down.

kevinsawicki:
Closing this out since it hasn't any movement since the last round of feedback, please comment if you still plan on making further progress on this one.

PR #275, opened 2014-11-11
Add theme color, contrast, and saturation settings

kevinsawicki:
Initial work towards making themes adjustable by making certain Less variables defined by config settings and used in themes.

theme-settings

@simurai couple questions for you:

* What do you think the preset colors should be?
* What do you think the preset contrast percentages should be?
* What do you think the preset saturation percentages should be?

Depends on https://github.com/atom/atom/pull/4153

### DIFF
--- a/lib/themes-panel.coffee
+++ b/lib/themes-panel.coffee
@@ -44,17 +44,17 @@ class ThemesPanel extends View
           @div class: 'form-group', =>
             @label class: 'col-sm-2 col-lg-2 control-label themes-label text', 'Saturation'
             @div class: 'col-sm-10 col-lg-4 col-md-4', =>
-              @div class: 'theme-saturation theme-saturation-1', 'data-saturation': '20%'
-              @div class: 'theme-saturation theme-saturation-2', 'data-saturation': '10%'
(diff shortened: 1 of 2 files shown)


smlombardi:
One thing I did in my Stylesheet is add a different font size for retina displays/non-retina.  I work on a Thunderbolt display and then disconnect my MBP and work on the built-in retina.  Sometimes font sizes that look great no one display are too small on the other.  Just a thought.

braver:
How does this relate to https://github.com/atom/atom/pull/4153 which was recently deemed unsuccessful and therefore closed? Sounds like this is not going to happen without a major overhaul?

kevinsawicki:
> How does this relate to atom/atom#4153 which was recently deemed unsuccessful and therefore closed? Sounds like this is not going to happen without a major overhaul?

Yeah, basically this isn't going to happen in its current form.

@simurai and I are still discussing different ways to make themes configurable and hope to ship it in some other form in the future.

braver:
Thanks, that clears up the roadmap for my theme. Definitely interested in hearing what you guys come up with though. It's a tough nut to crack for sure.

PR #315, opened 2015-01-08
Fix unreadable grey buttons in package list

lachlanjc:
In the new package view, I noticed that the buttons were almost unreadable. Some CSS from atom/atom (specifically [this line](https://github.com/atom/atom/blob/master/static/icons.less#L10)) was interfering.

Before:
screen shot 2015-01-07 at 10 31 11 pm

After:
screen shot 2015-01-07 at 10 33 10 pm

### DIFF
--- a/stylesheets/available-package.less
+++ b/stylesheets/available-package.less
@@ -201,7 +201,6 @@
       }
 
       .author {
-        color: @text-color-subtle;
         margin-left: @component-padding;
         &:focus {
           outline: none;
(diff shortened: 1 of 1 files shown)


simurai:
Thanks for the PR but this should be changed on a theme basis. It looks like you're using the Seti theme, right? There is already a PR https://github.com/jesseweed/seti-ui/pull/72 that sould fix it. /ping @jesseweed

As a temp fix, in "Menu > Open Your Stylesheet" you could add this:

```css
.settings-view .btn {
  color: #fff;
}
```

lachlanjc:
Oh sorry, completely didn't realize that. Hope it gets fixed soon.

PR #320, opened 2015-01-08
use min-width to force minimum width to allow @component-padding in rem

### LINKED ISSUE #319: don't expect variables from ui-variables.less to be in px units
available-packages.less expects `@component-padding` to be in pixels, at line 223:

```
.status-indicator {
  padding: 0;
  width: max(4px, @component-padding/2);
```

If `@component-padding` is in something else (`rem` as it is in [my theme](https://github.com/braver/isotope-ui)) the entire stylesheet breaks.

Can this be styled without this requirement?

braver:
`min-width` accomplishes the same as the `max()` calculation: make the status indicator halve the size of `@component-padding`, but make sure it's at least 4px. Doing it like this leaves it up to the client to figure it out render-time, which allows `@component-padding` to be in something else than pixels (e.g. rem).

fixes #319

### DIFF
--- a/stylesheets/available-package.less
+++ b/stylesheets/available-package.less
@@ -220,7 +220,8 @@
 
       .status-indicator {
         padding: 0;
-        width: max(4px, @component-padding/2);
+        width: @component-padding/2;
+        min-width: 4px;
         pointer-events: none;
         transition: background .4s;
         background: @background-color-success;


braver:
Just as a note, I understand most themes will be designed in pixels, and independence of specific units may not be upheld forever if everyone expects everything to be in pixels. However, I love the power, simplicity and potential beauty of rem-based layouts and this is the only line of stylesheet in all packages (core and otherwise) that I tested where a rem based variable results in problems. Really, the only thing you cannot do with a rem value is compare it to a ps value and it'd be awesome if I could keep my theme in rems so users can change font-sizes at will.

simurai:
:+1: The concept of `rem` based units for the UI part is great. I'll advocate it part of making themes configurable. Or maybe even as a setting in core, like the editor font size. @lee-dohm is also a fan. :grin:

lee-dohm:
Yes, yes I am :grinning:

braver:
Awesome! :trophy:

PR #325, opened 2015-01-09
Add color settings

kevinsawicki:
Depends on https://github.com/atom/atom/pull/4930

screen shot 2015-01-08 at 4 46 07 pm

### DIFF
--- a/lib/settings-panel.coffee
+++ b/lib/settings-panel.coffee
@@ -33,7 +33,7 @@ class SettingsPanel extends View
 
     @appendSettings(namespace, settings)
 
-    @bindCheckboxFields()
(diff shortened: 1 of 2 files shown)


kevinsawicki:
@simurai any feedback on this? Do you think we should explicitly style these to be the same on all platforms? (I'm assuming that is possible like the checkboxes).

simurai:
Yeah.. I styled it a bit. Looks now like this:

screen shot 2015-01-10 at 10 37 55 pm
screen shot 2015-01-10 at 10 38 37 pm

kevinsawicki:
@simurai Nice, thanks!

Issue #349, opened 2015-01-21
Use native controls in core


simurai:
The checkboxes have gone through some changes recently: Native > toggle switches > custom styled. After some discussions, the checkboxes will be changed back to native. Mainly because of this reason:

> Atom core has no opinion about look and feel. It just has enough styling so that it's functional. The rest is up to the themes to define.

It's also easier to add styling on top of core, instead of trying to reset some fancy styles coming from core. The settings-view will still adapt to themes, but only things like background color or text color that come from a theme's `syntax-variables.less`. Custom checkboxes, color picker would go too far.

#### What does this mean for themes?

Changing the checkboxes back to native will make them look a bit strange in dark themes. So themes will have to custom style them on their own (if they want). I'll move the current custom checkboxes into the [One](https://github.com/atom/one-dark-ui) themes, if somebody wants to copy them.

simurai:
Closing.. see discussion on #350

Issue #318, opened 2015-01-08
Centered layout is awkward on wide displays


mnquintana:
I love what you've all been doing with the Settings UI so far! But I did notice that the new centered single column looks a little awkward on wide displays:

![screenshot from 2015-01-08 09 39 16](https://cloud.githubusercontent.com/assets/823545/5664131/e9a5bb94-971b-11e4-8efb-1ba143e39758.png)
![screenshot from 2015-01-08 09 38 35](https://cloud.githubusercontent.com/assets/823545/5664003/a42146b6-971a-11e4-911d-421f518f3356.png)

It seems like it might work better with multiple columns to better take advantage of all that negative space?

postcasio:
Yeah, it looks even more awkward on a 2560x1440 display. I added this to my stylesheet to get it into two columns, and a little wider:

```less
.settings-view {
	.section-container {
		max-width: 50%;
	}
	.package-container .row {
		width: 50%;
		float: left;
		padding: 0 0 0 (@component-padding / 2);
		&:nth-child(2n+1) {
			clear: both;
			padding: 0 (@component-padding / 2) 0 0;
		}
	}
}
```

philipgiuliani:
I also noticed this today. I would like to have it left-aligned and not centered.

thedaniel:
For right now, centering is a good compromise for varying window sizes IMO, but this could definitely be better. 

I am strongly :-1: on multiple columns. I had multiple columns in the original prototype for this, but removed them because they make lists of related content difficult to scan quickly, which I think is the most important usage pattern of this view.

thedaniel:
I'm not going to stand in the way of a nice left-aligned list though if someone can come up with one :)

mnquintana:
@thedaniel I'm not so convinced that multiple columns would make it too difficult to scan content (for the Packages pane anyway), as long as the columns are kept to a reasonable number (ie. 2 or 3). I actually found the previous incarnation of settings more readable because with the 2 column layout scanning packages (in the Featured section anyway) didn't entail having to scroll forever.

simurai:
Ya, I agree it looks a bit odd on very large screens. Personally I find it less problematic because I either have 2 panes open or two windows side by side. But could be just me.

[...]

PR #350, opened 2015-01-21
Native controls

simurai:
Changes custom styled controls to native. See #349

### TODO:

- [x] Checkbox
- [x] Color Picker
- [ ] Select?
- [ ] Button?


simurai:
I was gonna push this straight to master, but then thought making a PR.

Also because a question came up: Should the `<select>` element also be native? Currently it looks like this:

screen shot 2015-01-21 at 3 10 08 pm

But if it should be consistent with the checkbox, then we should change it as well to native so that it looks like (on OS X):

screen shot 2015-01-21 at 3 12 15 pm

And should we go even further? What about `<button>`s? /cc @orderedlist

lee-dohm:
I think that if we're going to make a change like this, now is the time to do it. If there really is a concern that baseline Atom should be all native to make things easier on theme authors ... then let's do it :laughing: I'll post a link to this on Discuss to help give people fair warning and let them weigh in.

simurai:
I think this PR also depends if/when we want to move forward with taking the settings out of the tabs and put it into a "side panel". See the "Modal" section: https://github.com/atom/settings-view/issues/318#issuecomment-69690084

Because in a "side panel", the settings would be more its own thing and would not use a theme's color variables, but rather have its own default look. So theme creators wouldn't have to bother about the settings at all. Well, if they want, they could still override it, but it would be opt-in.

simurai:
My opinion about this keeps flip-flopping by the day. :stuck_out_tongue_closed_eyes: Some arguments to **not** merge this PR:

1. It's not for sure that most themes will start to custom style the native controls.
2. As mentioned above, it's hard to draw the line. If it should be consistent, then `<select>`, <button> and maybe more should also be native.
3. In case Atom adds a new element, like a radio or number input, we have to ask theme creators again to custom style those too.

I'd love to have a clear rule that core only uses native elements, but I think it's not practical at the moment. I'd suggest we wait to consider this until a 2.0.

mnquintana:
FWIW, I tend to avoid community themes (in pretty much any program) because they are almost always incomplete compared to the default themes. There's always some part of the program that ends up totally unstyled or poorly styled which to me makes community themes range from unpolished to unusable. If this PR is adding another thing for themers to have to remember to style, I'd be concerned about the end results of that :grimacing:. I feel like it might not be so bad for an Atom core package to have a sensible default styling for controls that themes can override as they wish.

braver:
I would definitely prefer controls you don't have to style if you don't want to. While I would usually prefer custom controls (my first reaction to this was really positive), they're simpler and easier to use, it's hard to create a smooth and consistent UI with them without going fully platform compliant. You give up all control over alignment, basically have to use default buttons and color schemes as well.... That might be ok if you're using a platform specific development kit, but since we're developing in Chrome and multi-platform I'm afraid Atom's default might end up looking bad in anything but OSX. If you're worried about theme creators, I would prefer some smart variables. Perhaps there's nothing wrong with the current situation and you shouldn't try to 'fix' it?

orderedlist:
> it's hard to create a smooth and consistent UI with them without going fully platform compliant. You give up all control over alignment, basically have to use default buttons and color schemes as well

I'm not sure this is entirely true. And we're going to be ensuring that Atom, in general, ships well on each platform, so having a non-broken settings view should likely go right along with that step. 

I know that it's super-easy to customize these settings, specifically because we're only targeting one browser... but I'm not sure that's enough reason to buck the HIG. Remember, we're making desktop software here. It's not the wild-west of the web. There's traditions and guidelines for interactions. Following them isn't going to be the easier route, but I might argue it's the better one.

braver:
> There's traditions and guidelines for interactions. Following them isn't going to be the easier route, but I might argue it's the better one.

True, I would argue the same. I'm just flip-flopping along with simurai: either way has much going for it and as long as you guys choose something you're comfortable with (while knowing all trade-offs) would be understandable. I've made similar decisions before and have gone both ways depending on the situation.

simurai:
> I'm not sure that's enough reason to buck the HIG.

@orderedlist I think you mean not just the checkbox, or select to be native, instead the whole settings should look like it's from a native app, right?

Hmmm.. Currently the settings are "inside a tab". If it's inside a tab it should look like it's part of the theme. Meaning the settings should adapt to colors of a theme and use the `ui-variables.less`. That's what it currently does. So as of today, themes have priority over OS. Having a totally custom looking app that has a native looking part attached into the middle of the UI, feels a bit odd.

Now, if the OS should have priority and the settings should look as close as possible to native, then we should pull the settings out from the tabs and move it into an own window on OS X and Linux and maybe on Windows too or as a side panel. Probably not doable with CSS alone, and would need different HTML/JS. Also on Linux there are lots of different flavors that all have a different look. So we might end up having to create and maintain 5-6 versions of the settings. I'm not against the idea, just not sure if it's high on the priority list, thinking of the amount of work that would be needed. Even the current settings aren't "done" yet and could still use lots of improvements.

So if everyone agrees that as long as the settings are inside a tab, it should look like it belongs to the theme, then I would suggest this:

1. We close this PR without merging.
2. In the Unity theme (that aims to look like native), we unset all custom styled controls. Not just checkbox, but also selects, button, inputs etc. It would still be inside a tab, but should fit in with the rest of the UI.

Once/if the settings get moved out of the tabs, could be an own window or side panel described [here](https://github.com/atom/settings-view/issues/318#issuecomment-69690084), then we can pick up this discussion again.

braver:
> Having a totally custom looking app that has a native looking part attached into the middle of the UI, feels a bit odd.
> ... So we might end up having to create and maintain 5-6 versions of the settings...

I guess that sums up my argument against: if you're going to stick some native ui into it, it's not going to work. It's a good ambition to go as native as possible and I would agree Atom should move that way, but you have to weigh it against the cons and do it well if at all. If you're going to keep it as is, it's not wrong and Atom isn't on its own.

simurai:
Ok, I think I'll close this PR again without merging. I made a new PR to add native checkboxes in the Unity theme: https://github.com/orderedlist/unity-ui/pull/54

@orderedlist I hope that is somewhat acceptable, at least for now with the settings in the tabs.

Issue #328, opened 2015-01-09
There's no easy way to open Atom.io from package cards


izuzak:
_From @gfarrell on January 9, 2015 12:14_

When trying to find out more about a package, in the *Install* view of the *Settings* page in Atom `0.169.0`, clicking the title does nothing, but clicking the author opens the author page in a browser.

_Copied from original issue: atom/atom#4947_

zephraph:
+1

This is slightly annoying.

thedaniel:
This is not a bug as described. See my comments on #351. Updating the title of this issue to clarify.

mnquintana:
I think this was resolved in #373

Issue #330, opened 2015-01-09
Indicate when links are external (ie. when they'll open in a browser)


mnquintana:
Sort of related to #328 – it would be useful if there were a trailing icon (much like Wikipedia) on every link in Settings view indicating that the link would open in a browser, and not in Atom internally. Currently the icon I have in mind is being used for the "View Code" button on a package's settings.

![screenshot 2015-01-09 09 23 19](https://cloud.githubusercontent.com/assets/823545/5680919/30fc0ba4-97e1-11e4-968b-c47611cae5e1.png)

simurai:
I liked the initial idea, but after trying it out, I'm not sure anymore.

![ext-links](https://cloud.githubusercontent.com/assets/378023/5757882/24c5d33e-9d03-11e4-98d7-4795ae3696b6.png)

It adds a lot of extra "visual noise". And having icons left AND right looks a bit strange. I think on websites like Wikipedia it's great because links have no other icons. I wonder if there is another, more subtle way?

simurai:
> I wonder if there is another, more subtle way?

Maybe just an underline when `hover`ing? Most links already do that and for the "buttons" it would like this:

![screen shot 2015-01-15 at 10 18 06 pm](https://cloud.githubusercontent.com/assets/378023/5758014/8592cdce-9d04-11e4-9327-aecd800fef9e.png)

mnquintana:
> It adds a lot of extra "visual noise". And having icons left AND right looks a bit strange. I think on websites like Wikipedia it's great because links have no other icons. I wonder if there is another, more subtle way?

Hmmm, I'm thinking the external link icon might be too big relative to the text (and the left icons). Maybe try making it smaller? I think part of the reason it adds more noise is that the stroke weight of the icon is kinda heavy, so I'm thinking that a smaller size would help compensate for that. It also is much less distracting to my eyes on the `.author` link than the `.repo-link` link, probably because of the color.

Other options I can think of:
- Maybe the icon would be hidden unless `hover`ing? But no I think that would defeat the purpose of predictability (ie. you should know if a link is external before you even `hover`)

> Maybe just an underline when hovering? Most links already do that and for the "buttons" it would like this:

That might be a little too subtle. I'm not sure it would be obvious enough that that would indicate an external link (I feel like it might just come off as an odd aesthetic design choice).

lee-dohm:
[...]

Issue #284, opened 2014-12-16
Unify core and package settings


thedaniel:
This came up in the discussion around #277 - it would be interesting to have the Settings pane contain all the package settings as well, and use a fuzzy-finder to filter down to what you need. I would like to keep the individual package meta / settings pages still, to allow something like #90 to be supported down the road, but there seems to be value in a single, global, searchable settings page. Failing that, at least some kind of global search that then allows the relevant package settings to be found and opened (a la the OS X System Preferences search box)

cc @maxbrunsfeld @simurai

davefp:
:+1:

I've so far found it difficult to locate particular settings since I first have to know/find the package containing it. Fuzzy search across all settings would be rad.

danielo515:
I have to agree with this request.

fulldecent:
To demonstrate best practice, attached are Mac and Chrome screenshots

![screen shot 2015-02-18 at 11 39 18 am](https://cloud.githubusercontent.com/assets/382183/6251609/d7214684-b762-11e4-8b9f-f798b27b15a9.png)
![screen shot 2015-02-18 at 11 38 42 am](https://cloud.githubusercontent.com/assets/382183/6251608/d720c66e-b762-11e4-9712-f121a2748a25.png)

rbu:
Is this issue "just" for a global page for all settings, or *also* for search in all (plugin) settings? Or is there/should there be a separate bug for that?

mnquintana:
@rbu It's really for both, although this may end up getting implemented in multiple PRs.

rbu:
@mnquintana, thanks for the information. Just to weigh in: I care for "search in all settings" much more than "all settings in one dialog". Thanks you for this!

Zireael07:
Both search in all settings and show all settings in one place should happen. Sooner rather than later.

thedaniel:
> Sooner rather than later.

Right now we're working on the issues [in this list](https://github.com/atom/atom/issues/7995), but this is could be plausible to attempt in the fall. If anyone is interested in working on it before then I think a good place to start is to wireframe out how the search interaction might work in the settings view and think about how the current views in settings might have to change. (I think this is more of a UX problem than a matter of writing code)

chav-n:
[Included from atom/atom#8514 for reference, minor editions]

[...]

PR #351, opened 2015-01-21
Add link to package name

zephraph:
Fix for #328.

### DIFF
--- a/lib/available-package-view.coffee
+++ b/lib/available-package-view.coffee
@@ -25,7 +25,7 @@ class AvailablePackageView extends View
 
       @div class: 'body', =>
         @h4 class: 'card-name', =>
-          @a outlet: 'packageName', href: "https://atom.io/packages/#{name}", name
+          @a outlet: 'packageName', name
         @span outlet: 'packageDescription', class: 'package-description', description
 
       @div class: 'meta', =>
@@ -74,7 +74,10 @@ class AvailablePackageView extends View
       @parents('.settings-view').view()?.showPanel(@pack.name, {back: opts?.back})
 
     @packageName.on 'click', =>
-      @parents('.settings-view').view()?.showPanel(@pack.name, {back: opts?.back})
+      if @installButton.isVisible()
+        shell.openExternal("https://atom.io/packages/#{@pack.name}")
+      else
+        @parents('.settings-view').view()?.showPanel(@pack.name, {back: opts?.back})
 
     @enablementButton.on 'click', =>
       if @isDisabled()
--- a/lib/available-package-view.coffee
+++ b/lib/available-package-view.coffee
@@ -25,7 +25,7 @@ class AvailablePackageView extends View
 
       @div class: 'body', =>
         @h4 class: 'card-name', =>
-          @a outlet: 'packageName', name
+          @a outlet: 'packageName', href: "https://atom.io/packages/#{name}", name
         @span outlet: 'packageDescription', class: 'package-description', description
 
       @div class: 'meta', =>


thedaniel:
This view is used in a lot of different places, including list views. If we have this href set everywhere, if you click a package name in a list of packages, it will *both* open your browser and navigate within settings view to a package detail page (InstalledPackageView).

screen shot 2015-01-21 at 6 11 13 pm

I think a better thing to do would be to detect whether the AvailablePackageView is inside an InstalledPackatgeView or initialize it with a flag, and have a conditional in the click hander: https://github.com/atom/settings-view/blob/808cd73d72b98dab83ca1ee139f967cffc47cd43/lib/available-package-view.coffee#L76

zephraph:
All right, I'll try that out

zephraph:
I looked at a few different options. Checking if the available package view was inside the installed package view wasn't enough. Because the install view now shows already installed packages as well as packages not installed yet. The common link here is whether or not the package is installed. 

I made the simplest changed possible: check if the install button is visible. If it is, open the external link. If not, open the package view. I realize this may not be the "best" way to do it. 

An alternative would be to add a flag on the different events (when installing, installed, uninstalled, etc) that would determine how the on package name clicked event was handled. That would decouple the traversal from the visibility of the install button, but it would also add more code. 

If you would rather that method let me know and I can implement it. If you would rather a different method then I'm all ears. I'd just really like to get this fixed. It's a feature I used a lot in the older versions.

zephraph:
@thedaniel, would you mind reviewing this when you get the chance? Thanks!

thedaniel:
I've changed my mind on this - I think it's potentially confusing to have the same control in the same context (package card) have two different actions. Consistency is a virtue in UI. If it's important to visit Atom.io from the card we should add another element, perhaps an octicon next to the package name, something else in the button row, etc. 

cc @simurai in case he has any UI ideas.

Sorry that I encouraged you to implement this and then rejected it -- I guess I needed to stew on it a bit more.

thedaniel:
Also, for what it's worth I do think we should support easily opening atom.io from the card because the metadata there is probably always going to be a few revisions ahead of the settings view in terms of completeness etc.

thedaniel:
We could also consider trying to have some kind of visual consistency around what controls open the browser and what causes the settings view to navigate somewhere.. given that the package name and the author are both normally styled links but do two different things.

zephraph:
I was simply trying to match the functionality that use to be there. 

It isn't even that important to pull up atom.io from the card. I just expect to be able to click on the package title and get more information about that package so I can decide whether or not to install said package. 

Right now I have to open atom.io manually and look up the package. If you want consistency, what about creating another view that renders the README.md in a view similar to the single package view. 

Here's an example mockup:
picker

This is consistent with the current implementation and doesn't add clutter to the package cards. It also gives the users the ability to see what the package actually does. 

What do you think @thedaniel?

simurai:
Showing the `README.md` would be an option. Or do this:

1. Remove the underline on `:hover` of package names. So that it doesn't look like you can click it.
2. Add a "View on Atom.io" button infront of the "Install" button. It's the same as in the bottom row of a package detail page. Like this:

screen shot 2015-01-28 at 2 05 37 pm

thedaniel:
I'm down to show the README in any case outside of a link solution, it just wasn't a big priority for me right now. Happy to accept a PR and/or put it on the ol' todo list though. 

--
Daniel Hengeveld
(sent from my phone)
http://about.me/thedaniel

simurai:
I'm not sure about "show the README" inside the card in the package settings, like in the above mockup. I think that page should only be accessed after installing. But there could be a README button that would open a new tab. Like this:

screen shot 2015-01-28 at 5 33 54 pm

screen shot 2015-01-28 at 5 42 14 pm

Note that the screenshot is taken with a theme that styles the `markdown-preview` package. Most themes will have a white background.

So yeah.. I would say either:

1. the external "View on Atom.io" link
2. or "README" as a new tab

@zephraph Any preferences? My vote tends to go to the external "View on Atom.io" link.

zephraph:
Well, there are positives and negatives to both approaches. 

First off, your alternatives are more concise and I like that. The issue is just taking that approach _alone_ makes for interface inconsistencies of its own. When I install the package, the name is clickable. It does something... but when I don't have it installed the package name does nothing? It's slightly confusing because the views are essentially the same just with different buttons visible. So one approach would be to completely make the package name non-clickable. Technically the settings button is there so it's really redundant functionality, right?

So my position is this. If it's clickable, it takes you to a new view which is essentially what I mocked up. The thing about that is though, that's not the settings view. It'd be a new "package preview" view or something. Some of the links that are listed on there are definitely unnecessary but I left them in there because I was lazy. Point is, this isn't the settings view, just a view to provide information about the package and a quick way to navigate to all the important stuff related to the package. One of the biggest reasons I like this is because it gives consistent behavior between viewing an already installed package and a not-yet-installed package _and_ it doesn't open an external browser. 

As a retort to myself this will take longer to implement and it's not necessarily as visible as clicking an 'Open on Atom.io' button as @simurai mocked up. Plus it would mostly eliminate the need for anyone to use atom.io/packages if they had Atom installed. That being said, using something for the sake of using it really isn't a good design practice. 

I really like clickable packages names though. It's all about how close the package name is to the settings-view navigation. Removing it isn't a deal breaker, but I still prefer having the option to use it. 

So here are the options as I see it

Option 1:
Completely disable clickable package names on all views and add the open on atom.io button

Option 2:
Add a package preview view and have that open when clicking on an uninstalled package

Option 3:
Do option 2 + add the atom.io button. 

@thedaniel, @simurai thoughts comments?

mnquintana:
This is sort of related to #330, #328, but a few random thoughts in no particular order:
- Links that open in a browser (external links) should have some kind of visual indicator so you know what to expect whenever you click a UI component in settings-view - this might help mitigate expectations for users who click the package name and expect to be taken to atom.io
- I agree that there really isn't an easy way to get to atom.io/packages/<package-name> in the package card, which seems to be more of an issue when you're in the Install pane
  - This also seems to beg the question, how much of your package research / updating / installation workflow should be spent in Atom, and how much on atom.io? Personally, I think too much context-switching is a Bad Thing, and can much too easily break your flow, so I'd be in favor of having to bounce to atom.io as infrequently as possible.
- I like the idea of having a "View on Atom.io" link, but it seems really obtrusive as a button like that next to Install. I feel like I'd rather it be a clearly-defined icon or something (maybe with a tooltip on hover) next to the package name.
- I really like clickable package names too, and I think they should stay the way they are more or less (pretty much necessary for #355)
- It seems to me like "Install cards" have different use cases than installed package cards - for instance, if you're in the Install pane and looking at a package, you probably want to learn more about it, hence an easy way to get to the README seems important.
- @zephraph I think I like option 3 the most, as long as the atom.io button is less obtrusive. Otherwise, option 2, although that would be a lot more work.

simurai:
> @zephraph that's not the settings view. It'd be a new "package preview" view

It would look similar, but only have the things that make sense. ok, yeah.. I see.

Option 1, 2 and 3 could also be done in steps. Meaning do option 1 as a first step, since there is less effort involved and would already improve the current state. Then once that is launched, we can see how option 2 would work out and then decide about keeping the "Open on Atom.io" button or not (option 3).

mnquintana:
@simurai The only issue with option 1 is that it would make #355 impossible :confused:

simurai:
> @mnquintana This also seems to beg the question, how much of your package research / updating / installation workflow should be spent in Atom, and how much on atom.io?

If it's possible to hack an "Install" button onto atom.io, that would be amazing. Doing discovery inside Atom is ok too, but you can't easily share or link to a package.

> I like the idea of having a "View on Atom.io" link, but it seems really obtrusive as a button like that next to Install. I feel like I'd rather it be a clearly-defined icon or something (maybe with a tooltip on hover) next to the package name.

Hmm.. I probably would find an icon next to the package name a bit obtrusive. I know, there is the "external link" issue again #330 :innocent: But as a button group the two actions would be together. Like a cross-road: "Do you want more infos or go ahead and install it?"

simurai:
> The only issue with option 1 is that it would make #355 impossible :confused:

Hmm.. yeah. Well, option 1 could also be done by keeping the name clickable and just "disable" it for uninstalled cards. Let's call it 1B. Personally I think clicking the name is somewhat a hidden feature. More like a shortcut if you know about it, but not the obvious way I would expect the settings to be found.

zephraph:
> If it's possible to hack an "Install" button onto atom.io, that would be amazing.

I really like this and I'd love to discuss it more, but I think it's outside the scope of this PR. The scope of this issue has changed enough that I feel like I should go ahead and close this PR. I'm going to hash out option 2 this weekend and submit a new PR. Then, after everyone's had a chance to play with it, we'll come back to this discussion.

simurai:
> I'm going to hash out option 2 this weekend and submit a new PR.

Great. Let me know if you need some help with styling.

Just in case and maybe as a quick fix for the current situation, I made a PR for option 1B (still needs specs): #360

mnquintana:
> @simurai: Personally I think clicking the name is somewhat a hidden feature. More like a shortcut if you know about it, but not the obvious way I would expect the settings to be found.

Too bad we don't have any data about that – looks like [metrics wouldn't know](https://github.com/atom/metrics#collected-data).

> Hmm.. I probably would find an icon next to the package name a bit obtrusive. I know, there is the "external link" issue again #330 :innocent: But as a button group the two actions would be together. Like a cross-road: "Do you want more infos or go ahead and install it?"

Heh fair enough :stuck_out_tongue_closed_eyes:, it was only a half-baked idea. That's a good point about it being a crossroad – I think my brain is just not used to seeing another button there (with so many words) so I'll just give it time.

thedaniel:
> If it's possible to hack an "Install" button onto atom.io, that would be amazing.

This is a thing that has been an open item on atom/atom for a while. The latest version of the issue is here: https://github.com/atom/atom/issues/5262

I'd personally like to get it in sooner rather than later but the 1.0 feature list is already quite large so it unfortunately is scheduled for "when someone has time to get to it".

> I really like this and I'd love to discuss it more, but I think it's outside the scope of this PR. 

This is correct.

thedaniel:
> [clicking the package name is ] more like a shortcut if you know about it, but not the obvious way I would expect the settings to be found.

Maybe my perspective is tainted because I implemented it, but to me clicking on the name or the card body *is* the obvious way to navigate forward.

zephraph:
@thedaniel I kinda agree. It's intuitive to me to interact with the package card in that way.

mnquintana:
@thedaniel I agree too. I actually would expect the settings button to jump you to the Settings section of the package detail view (like a `/#something` link).

simurai:
There is also #284. Having all package settings under the main "settings" menu.

Then the current "package settings" page is not needed and could be replaced by the "package preview" page. Or it would be more like "package info". So clicking on the "Settings" button would go to `settings/#package-name` and clicking on the name would go `packages/#package-name`.

thedaniel:
I don't know how I feel about removing the package settings from the detail page even if they are all on the main settings page as well..

simurai:
Ohh.. wait. I agree, better to keep it in both places. I forgot about the keybinds of packages.

thedaniel:
@simurai @zephraph I started a PR with @jlord trying to pull in all of the UI considerations around the package card in #373 - I think we can address this PR as well as #360 in one holistic attempt.

PR #360, opened 2015-01-29
Add link to atom.io when a package is not installed

simurai:
This PR is option 1B from the discussion in #351. It maybe just be a stop gap solution. It...

1. Adds an "View on Atom.io" button next to the "Install" button
2. Disables the package name link (no underline when hovering)

io

#### Todo:

- [ ] Add specs?

I guess it would also need some specs? I could try to add it, but not sure if it's a good idea. :grin:

### DIFF
--- a/lib/available-package-view.coffee
+++ b/lib/available-package-view.coffee
@@ -137,6 +137,7 @@ class AvailablePackageView extends View
       unless error?
         @updateEnablement()
 
+        @packageName.removeClass('is-disabled')
         @learnMoreButton.hide()
         @installButton.hide()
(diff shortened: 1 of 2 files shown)


lee-dohm:
:clap: Very cool ... I like it!

thedaniel:
I was pairing with @jlord on some changes related to this and we really like just having the package name always link out and have the in-app navigation come from clicking on the card body - check it out in  #373

simurai:
ok, yeah, that would make it more consistent with the author name (both beeing a link with underline).

I guess we can close this one again.

PR #369, opened 2015-02-06
Support URIs for each of the panels

benogle:
Support things like this

```coffee
atom.workspace.open('atom://config')
atom.workspace.open('atom://config/themes')
atom.workspace.open('atom://config/install')
```


benogle:
WTF builder. It passes on my machine. :-1:

PR #370, opened 2015-02-06
Normalize order of displayed packages

sedabull:
This is not a major issue, just a minor annoyance that I've noticed over the past few weeks. The order in which installed packages are displayed seems unnatural to me for the following reasons:
* `dev-packages` generally make up the smallest sub-section of installed packages, they should therefore be displayed first
* `dev-packages` are also adjusted more often than `community` or `core` packages, by people trying to tweak them before a release
* `dev-packages` take precedence over `community-packages`, so they should be listed before, not after the `community-packages`
* `core-packages` are almost never going to change, and will therefore be needed least often and should be displayed last

Like I said, not a big deal, it's just another one of those minor details that, if adjusted, will make the user experience just that much better!

### DIFF
--- a/lib/installed-packages-panel.coffee
+++ b/lib/installed-packages-panel.coffee
@@ -22,6 +22,13 @@ class InstalledPackagesPanel extends View
 
           @div outlet: 'updateErrors'
 
+          @section class: 'sub-section dev-packages', =>
+            @h3 class: 'sub-section-heading icon icon-package', =>
+              @text 'Development Packages'
+              @span outlet: 'devCount', class:'section-heading-count badge badge-flexible', '…'
+            @div outlet: 'devPackages', class: 'container package-container', =>
+              @div class: 'alert alert-info loading-area icon icon-hourglass', "Loading packages…"
+
           @section class: 'sub-section installed-packages', =>
             @h3 class: 'sub-section-heading icon icon-package', =>
               @text 'Community Packages'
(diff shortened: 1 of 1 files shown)


simurai:
I assume if there are no dev packages installed, it would say "Development Packages (0)". So for most people, always having that at the top looks a bit odd.

Is it possible to not show the titles while loading and just if a group gets populated? Then yeah, the dev packages could be at the top.

Another option might be to make the groups collapsable. Then you wouldn't need to scroll down all the time. Or a sub-nav?

50Wliu:
Collapsable groups would be excellent.
I disagree though that the dev packages should be at the top.  The majority of users as @simurai said will not have any dev packages.  Those who do will know where to look for them.

sedabull:
I'm +1 on collabsable groups, though I don't think I'm the one to implement them. I respectfully disagree, however, about dev packages being displayed first. Atom promotes itself as being *the* "hackable text editor". A huge part of that is the ability to `apm link --dev` local projects. Displaying the developer packages first, even if most people won't have any, reinforces the idea that hackability is first and foremost what atom is all about.

thedaniel:
I think I agree with @simurai and @50Wliu on this one - I don't think we need to put an empty dev packages section at the top to make a point about hackability. I'd be down to hide it until you `apm link -d` something and *then* put it at the top though, because I think that if you *are* a package developer you probably spend more time adjusting package settings for your dev packages than anything else. I am +1 on collapsible sections and -1 on a subnav (too many navs on one screen make thedaniel crazy) unless they are the most minimal thing in the world.

thedaniel:
Similarly the community package section could have an interesting empty state, perhaps giving the user a link to package search with some helpful text if there's nothing installed.

thomasjo:
Basically everything @thedaniel just wrote :+1:

sedabull:
Very well, I'll concede that most people won't need the `dev-packages` section to be front and center, but I do think it should at least have an interesting empty state as well. Something like:
```
This is the section where all the packages you're developing locally appear.
To learn more about this awesome Atom feature click here!
```
Also, moving the `dev-packages` section to the top for people who do use it would be very much appreciated!

sedabull:
In the interest of reducing noise, I'm going to go ahead and close this. No need to keep a two month old PR open :smiley:

PR #377, opened 2015-03-02
Settings view allows installing package versions not supported by the current version of Atom

abe33:
When installing a new package, Settings View will allow package versions to be installed whose `package.json` files require a version of Atom newer than the one currently installed.

See atom/atom#5467 for details.

### DIFF
--- a/lib/package-manager.coffee
+++ b/lib/package-manager.coffee
@@ -4,6 +4,8 @@ _ = require 'underscore-plus'
 Q = require 'q'
 semver = require 'semver'
 url = require 'url'
+apm = require 'apm'
+request = require 'apm/lib/request'
 
 Client = require './atom-io-client'
 
@@ -117,6 +119,49 @@ class PackageManager
 
   getPackage: (packageName) ->
     @packagePromises[packageName] ?= Q.nbind(@loadPackage, this, packageName)()
+
+  # Request package information from the atom.io API for a given package name.
+  #
+  # packageName - The string name of the package to request.
+  # callback - The function to invoke when the request completes with an error
+  #            as the first argument and an object as the second.
+  requestPackage: (packageName, callback) ->
+    requestSettings =
+      url: "#{apm.getAtomPackagesUrl()}/#{packageName}"
+      json: true
+    request.get requestSettings, (error, response, body={}) ->
+      if error?
+        callback("Request for package information failed: #{error.message}")
+      else if response.statusCode isnt 200
+        message = request.getErrorMessage(response, body)
+        callback("Request for package information failed: #{message}")
+      else
+        if body.releases.latest
+          callback(null, body)
+        else
+          callback("No releases available for #{packageName}")
+
+  getLatestCompatibleVersion: (pack) ->
+    atomVersion = @normalizeVersion(atom.getVersion())
+    return pack.releases.latest unless atomVersion?
+
+    latestVersion = null
+    for version, metadata of pack.versions ? {}
+      continue unless semver.valid(version)
+      continue unless metadata
+
+      engine = metadata.engines?.atom ? '*'
+      continue unless semver.validRange(engine)
+      continue unless semver.satisfies(atomVersion, engine)
+
+      latestVersion ?= version
+      latestVersion = version if semver.gt(version, latestVersion)
+
+    latestVersion
+
+  normalizeVersion: (version) ->
+    [version] = version.split('-') if typeof version is 'string'
+    version
 
   search: (query, options = {}) ->
     deferred = Q.defer()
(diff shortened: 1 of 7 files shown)


abe33:
Actually it just hide the install button if the package's atom engine version isn't satisfy by the current Atom version. 

I think we should display some information to the user in that case, like a `text-danger` span with the expected Atom version and some explanation.
Regarding the update card, I believe it should follow the same pattern, being displayed as an update, but marked as unavailable and accompagnied with a version mismatch message. 

For the last point I think apm play a big role in how the settings view handle the package updates, so I'll need to make some tests with dummy packages.

kevinsawicki:
Should it just install the version that is compatible when one does exist?

Ajedi32:
> Should it just install the version that is compatible when one does exist?

Yeah, that's how APM behaves.

abe33:
> Should it just install the version that is compatible when one does exist?

Sure, it would be ideal. I just need to know how to retrieve the compatible versions of a package for a given version of Atom.
I'll take a look at how apm does that.

abe33:
So, now, when displaying a package that is not installed a request is made to the API using the `apm` module and copy of the `requestPackage` and `getLatestCompatibleVersion` from the install command.

If a compatible version is available the install button will install this version.

It still lack messages when the package can't be installed and changing the card's data with the one from the latest compatible version.

abe33:
Now install cards displays a message according to the availability of a compatible version. 

Find below two screenshots of the settings view for the two situations: 

has compatible version
does not have compatible version

As I'm not a native English speaker the wording is obviously open to changes.

Ajedi32:
Fixes abe33/uninstallable#1. :wink:

abe33:
I've merge back the recent changes in the package card. 

There's a minor change from the version before the merge. If the latest version can't be installed but there's a compatible version the `@pack` variable is no longer replaced with the compatible version but rather stored in a `@installablePack` variable instead. That way, when clicking on the card in the list to view the details, the version compatibility messages can still be displayed in the detailed view.

Let me know if there's anything I can do to improve this PR.

cc/ @thedaniel @kevinsawicki

abe33:
Bump @kevinsawicki @thedaniel @izuzak @simurai 

Anyone wanting to give some feedback on this?

lib/package-manager.coffee
```diff
@@ -4,6 +4,8 @@ _ = require 'underscore-plus'
 Q = require 'q'
 semver = require 'semver'
 url = require 'url'
+apm = require 'apm'
```
kevinsawicki: Wow, I didn't know this was possible, it definitely isn't intended to be on the require path.

This apm is meant for standalone use only since it is compiled as a command line utility and so its native modules are built against node and not atom shell.

kevinsawicki: I'd like to keep the settings view only interacting with apm via an external process so we can ensure apm exposes a useful enough API to be used.

What do you think about making this flow possible from an `apm` command such as something like `apm view --json --compatible 0.188.0` or somethign?

abe33: Sounds good to me. 
I'll open an issue on apm to support the `--compatible` option in the `view` command.

abe33:
The latest commit update the `PackageManager` class to add a `loadCompatiblePackageVersion` method that use `apm view --compatible version` command under the hood. Except from that change the behavior is still the same as the one I described above.

abe33:
Maybe this PR could have been merged in the latest release. 
If someone can take some time to review it.
cc/ @jlord @thedaniel @kevinsawicki

lib/package-manager.coffee
kevinsawicki: This and the `request` require can be removed now right?

abe33: Done :)

spec/installed-package-view-spec.coffee
```diff
@@ -2,7 +2,11 @@ path = require 'path'
 PackageDetailView = require '../lib/package-detail-view'
 PackageManager = require '../lib/package-manager'
 
+
```
kevinsawicki: :fire:

abe33:
Any plan to merge this PR? 

cc/ @kevinsawicki

joefitzgerald:
Bump @kevinsawicki. This impacts autocomplete-plus fairly frequently.

mnquintana:
@abe33 In the meantime, would you mind merging in master and fixing the merge conflicts?

abe33:
@mnquintana I'm on it :)

thedaniel:
Thanks @abe33 - this looks great to me. I'm going to merge.

PR #407, opened 2015-03-05
Enablement updates

jlord:
This adds a spec and fixes enablement styles (whether a package is enabled or disabled) which were going missing when navigating to `detail-package-view`.

Problem:
Disabled in this view
screen shot 2015-03-04 at 9 25 55 pm
Enabled in this view
screen shot 2015-03-04 at 9 25 58 pm

### DIFF
--- a/lib/package-card.coffee
+++ b/lib/package-card.coffee
@@ -95,7 +95,6 @@ class PackageCard extends View
         atom.packages.enablePackage(@pack.name)
       else
         atom.packages.disablePackage(@pack.name)
-      @updateEnablement()
       false
 
   detached: ->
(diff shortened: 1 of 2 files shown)


lib/package-card.coffee
```diff
@@ -141,6 +140,12 @@ class PackageCard extends View
         .removeClass('is-disabled')
 
   handlePackageEvents: ->
+    atom.packages.onDidDeactivatePackage (pack) =>
+      @updateEnablement() if pack.name is @pack.name
```
thedaniel: what's this `if pack.name is @pack.name` all about? (just curious)

thomasjo: `atom.packages.onDidDeactivatePackage` fires for all packages, so this line ensures the package card is only updated if the deactivated package is the same as the package for the card.

thedaniel: Ha that is totally obvious now that I look at it again - thanks!

jlord:
Good to go :grey_question:

thedaniel:
Ⓛ Ⓖ Ⓣ Ⓜ

jlord:
Woot! Thanks @thedaniel!

Issue #416, opened 2015-03-10
Failed to activate the minimap package


johnmaia:
[Enter steps to reproduce below:]

1. ...
2. ...

**Atom Version**: 0.187.0
**System**: Mac OS X 10.10.2
**Thrown From**: [settings-view](https://github.com/atom/settings-view) package, v0.184.0

### Stack Trace

Failed to activate the minimap package

```
At undefined is not a function

TypeError: undefined is not a function
  at Package.module.exports.Package.activateServices (/Applications/Atom.app/Contents/Resources/app/src/package.js:355:117)
  at Package.module.exports.Package.activateNow (/Applications/Atom.app/Contents/Resources/app/src/package.js:243:16)
  at /Applications/Atom.app/Contents/Resources/app/src/package.js:221:30
  at Package.module.exports.Package.measure (/Applications/Atom.app/Contents/Resources/app/src/package.js:167:15)
  at Package.module.exports.Package.activate (/Applications/Atom.app/Contents/Resources/app/src/package.js:214:14)
  at PackageManager.module.exports.PackageManager.activatePackage (/Applications/Atom.app/Contents/Resources/app/src/package-manager.js:465:21)
  at /Applications/Atom.app/Contents/Resources/app/node_modules/settings-view/lib/package-manager.js:226:29
  at exit (/Applications/Atom.app/Contents/Resources/app/node_modules/settings-view/lib/package-manager.js:43:16)
  at BufferedProcess.triggerExitCallback (/Applications/Atom.app/Contents/Resources/app/src/buffered-process.js:63:47)
  at ChildProcess.<anonymous> (/Applications/Atom.app/Contents/Resources/app/src/buffered-process.js:85:18)
  at ChildProcess.emit (events.js:119:17)
  at Process.ChildProcess._handle.onexit (child_process.js:1072:12)

```

### Commands

```
     -4:58.7 core:move-right (atom-text-editor.editor)
     -4:53.8 core:backspace (atom-text-editor.editor)
     -4:51.7 editor:consolidate-selections (atom-text-editor.editor)
     -4:51.7 core:cancel (atom-text-editor.editor)
     -4:50.2 core:backspace (atom-text-editor.editor)
     -4:49.5 core:save (atom-text-editor.editor)
     -2:49.4 remote-edit:browse (atom-pane.pane.active)
  2x -2:26.5 core:backspace (atom-text-editor.editor.mini)
  3x -2:25.2 core:confirm (atom-text-editor.editor.mini)
[...]

PR #423, opened 2015-03-11
Unload old packages from module cache.

mathiasvr:
fixes atom/atom#5940
Before this, old package versions would be loaded from the node.js module cache through require(), after doing a package update. This would sometimes produce errors since only the main module of the package was outdated.

### DIFF
--- a/lib/package-manager.coffee
+++ b/lib/package-manager.coffee
@@ -151,7 +151,10 @@ class PackageManager
       activateOnSuccess = not atom.packages.isPackageDisabled(name)
     activateOnFailure = atom.packages.isPackageActive(name)
     atom.packages.deactivatePackage(name) if atom.packages.isPackageActive(name)
-    atom.packages.unloadPackage(name) if atom.packages.isPackageLoaded(name)
+    if atom.packages.isPackageLoaded(name)
+      # remove old version from node.js module cache
+      delete require.cache[atom.packages.getLoadedPackage(name).getMainModulePath()]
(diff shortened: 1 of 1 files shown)


izuzak:
cc @jlord @maxbrunsfeld for :eyes: 

I just tried this out and it seems that it doesn't fix https://github.com/atom/settings-view/issues/416 and other similar issues.

mathiasvr:
@izuzak Just tried your instructions and indeed the issue still exists for the minimap package. I initially thought reloading the main module would fix the problem, but after some debugging it seems that all modules required through the main module is cached in the same context as well. I think the safest way to solve this is to remove all modules related with package from the cache, since any of them could change during the update.
I will try to fix this pull request, but maybe it would be better to patch unloadPackage https://github.com/atom/atom/blob/master/src/package-manager.coffee#L366
in atom instead?

mathiasvr:
Uncaching the modules makes the package reload correctly but creates a new issue, in this case the minimap package fails to load because a call to Document.registerElement tries to re-register an element which the old version already registered. So unless there is a way to reset all the 'setup' a package loads in, a new version is likely to cause conflicts. The simplest solution would probably be to require an atom restart before reloading any updated packages, any thoughts on this?

abe33:
It's a shame that no method to revert a call to `registerElement` was proposed in the custom elements specifications.

mathiasvr:
Closing this PR, since it only fixes a limited range issues.

PR #373, opened 2015-02-06
Package card and detail view UI improvements

jlord:
- [x] Link to atom.io from package title
- [x] Click on body takes user to settings (detail) view
 - [x] Remove 'settings' button as that's the current view
- [x] Remove `status-indicator` on bundled un-status'ed themes
- [x] Fix disappearing breadcrumbs #358 
- [x] Show breadcrumbs on installed packages settings view
- [x] Show PackageDetailView for packages that _aren't_ installed
  - [x] PackageDetailView should change to installed state if package is installed from it
  - [x] PackageDetailView should change to uninstalled state if package is uninstalled from it
- [x] Show readme on on (future) install package detail view :point_up_2: 
 - [x] Remove readme button?

Overlaps with: #360, #351

PR #429, opened 2015-03-13
Add package card hover state

mnquintana:
Now that #373 has landed, I recently noticed that clicking on the package card body now takes you to the detail view for that package, which is quite nice, but threw me at first. It wasn't clear that anything would happen from clicking the package card body.

This adds a hover state to visually indicate that the user can interact with the package card body. I considered changing the cursor to a pointer too, but I wasn't sure if that'd be too weird - let me know if you think that's worth adding too!

screencap

### DIFF
--- a/lib/package-detail-view.coffee
+++ b/lib/package-detail-view.coffee
@@ -20,7 +20,7 @@ class PackageDetailView extends View
   Subscriber.includeInto(this)
 
   @content: (pack, packageManager) ->
-    @div =>
+    @div class: 'package-detail', =>
       @ol outlet: 'breadcrumbContainer', class: 'native-key-bindings breadcrumb', tabindex: -1, =>
         @li =>
           @a outlet: 'breadcrumb'
--- a/styles/available-package.less
+++ b/styles/available-package.less
@@ -1,5 +1,6 @@
 @import "octicon-mixins";
 @import "ui-variables";
+@import "variables";
 
 .settings-view {
 
@@ -12,19 +13,19 @@
     font-size: 1.2em;
     border-radius: @component-border-radius*2;
     border: 1px solid @base-border-color;
-    background-color: lighten(@tool-panel-background-color, 8%);
+    background-color: @package-card-background-color;
     overflow: hidden;
     cursor: pointer;
 
     &:hover {
-      background-color: lighten(@tool-panel-background-color, 10%);
+      background-color: contrast(@package-card-background-color, darken(@package-card-background-color, 2%), lighten(@package-card-background-color, 2%));
     }
     &:active {
-      background-color: lighten(@tool-panel-background-color, 8%);
+      background-color: @package-card-background-color;
     }
 
     &.disabled {
-      background-color: lighten(@tool-panel-background-color, 5%);
+      background-color: darken(@package-card-background-color, 3%);
       .body,
       .avatar,
       .author,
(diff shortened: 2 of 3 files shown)


mnquintana:
It currently also shows on the package detail view, which I don't want – currently working on the best way to handle that.

50Wliu:
I'd like the cursor being changed to the 'clickable' one.

kevinsawicki:
Looks cool, what do you think @simurai ?

simurai:
Yes, I like the idea. I made a new PR https://github.com/mnquintana/settings-view/pull/1 on top of this one. How do you feel about:

- Lighten on hover, instead of darken. Since most themes do that for buttons etc.
- No hover for disabled packages. It makes disabled packages feel more "disabled". :smile: There is still the pointer cursor to let you know it's clickable.. see next point.
- Add a pointer cursor on the whole card. I also felt a bit weird about it, but after trying it out, I think it's ok. Also, because all the other things that can be clicked have some sort of hover state, so you still see that it's something else.

mnquintana:
So the only thing left I think is to remove the hover state when on the package detail view.

simurai:
This would work:

```less
.breadcrumb + .section .package-card {
  cursor: auto;
  &:hover {
    background-color: lighten(@tool-panel-background-color, 8%);
  }
}
```

But it's kinda fragile to rely on the `.breadcrumb` class to be a sibling of `.section`. Maybe better to add some other class in the detail view.

mnquintana:
> Maybe better to add some other class to the detail view.

That's what I was thinking – I'll finish that up tonight.

simurai:
Made a PR, in case: https://github.com/mnquintana/settings-view/pull/2

Called the class `.package-detail`.

mnquintana:
Alright, I think this is about ready!

thedaniel:
This does not work with light themes.

thedaniel:
(or, it may work but the 8% light isn't noticeable in one-light)

mnquintana:
@thedaniel I had a feeling that would be the case (that's why I started by darkening the background color on hover) – so I could either go back to darkening the bg in all cases, or add some Less logic to darken if the background is light, otherwise lighten if the background is dark.

thomasjo:
@mnquintana Well, darkening will also have similar problems with (very) dark themes. We need a less fragile solution. Perhaps some clever usage of LESS' [contrast function](http://lesscss.org/functions/#color-operations-contrast) in combination with `lighten` and `darken`? :thought_balloon:

mnquintana:
@thomasjo I actually didn't know about the contrast function – that's kind of cool. I was going to go with a conditional approach like:

```less
&:hover when (lightness(@tool-panel-background-color) < 90%) {
    background-color: lighten(@tool-panel-background-color, 10%); // lighten
}
&:hover when (lightness(@tool-panel-background-color) >= 90%) {
  background-color: lighten(@tool-panel-background-color, 5%); // darken
}
```

thomasjo:
Experiment with both, and see what yields the desired/best result. Big :+1: once this works better for both light and dark themes.

mnquintana:
The contrast function worked quite nicely – the hover color should now adjust based on the package card background color. Since we were using it heavily, I also defined the package card background as a variable – let me know if you think it would be wise to namespace the file name (like `settings-view-variables.less` or something).

thomasjo:
> The contrast function worked quite nicely – the hover color should now adjust based on the package card background color

Awesome!

> let me know if you think it would be wise to namespace the file name

Shouldn't be necessary since it's isolated within the package.

mnquintana:
Updated screencaps:

screencap
screencap

simurai:
:+1: Ok, I think this is good to :ship:

@mnquintana Thanks!

thedaniel:
Cheers!

Issue #432, opened 2015-03-13
search package doesn't auto-focus when jumping from command-palette


sinkingshriek:
It's not a bug but just a slight nuisance. 

On `Atom v0.187.0, Ubuntu 14.04`

Steps:-
- Open `command-palette`
- Search for `install package` and select `Install Packages and Themes`
 
The `Settings` tab show up with `Install` highlighted as expected but `Search packages` bar isn't auto-focused causing me to have to reach for my mouse and click on the bar. 
It'd be nice if it auto-focused like it does when I click on `Install`.

thedaniel:
Fortunately you can hit `Tab` twice and focus it! It should be focused immediately though, you're right.

sinkingshriek:
:smiley: I'll use that workaround for a while.

mnquintana:
I think I might've figured out what's going on here - `SettingsView.initializePanels` doesn't focus any panel even though it should when it calls `SettingsView.showPanel`. Once the panels are initialized, `focus` works as expected.

mnquintana:
I also can't seem to find where [`SettingsView.focus`](https://github.com/atom/settings-view/blob/master/lib/settings-view.coffee#L153) is ever used.

thedaniel:
@mnquintana it's possible it's unused, when we refreshed the package management UI stuff a lot was changed but not really properly refactored.

mnquintana:
Alright, now I think I have a better handle on it – it's a race condition. I tried `setTimeout`ing these lines for 2 secs:

https://github.com/atom/settings-view/blob/master/lib/settings-view.coffee#L72:
```coffeescript
@showPanel('Settings') unless @activePanelName
```

https://github.com/atom/settings-view/blob/master/lib/settings-view.coffee#L125
```coffeescript
@showPanel(name) if @panelToShow is name
```

And the search fields focused as expected.

Issue #394, opened 2015-02-25
Find should allow you to search within the Settings View


izuzak:
_From @benbalter on February 24, 2015 0:56_

I went to the packages list, looking to change a particular packages setting, and expected Command-F to allow me to search the list (thinking it was just an Atom view). Turns out that's not the case (I had to click into the search filter box, despite the native search dialog popping up).

It'd be awesome if either native search worked, or hitting Command-F hijacked the native functionality and dropped me in the filter dialog when listing installed plugins.

_Copied from original issue: atom/atom#5703_

izuzak:
_From @50Wliu on February 24, 2015 1:11_

I'd actually consider the find box popping up a bug (the best behavior would be to put focus into the filter dialog).

izuzak:
Related: https://github.com/atom/settings-view/issues/253

50Wliu:
Generalized the title a bit because fnr can be pretty useful in the package's settings as well.

damieng:
Closing in favor of a general settings search function.

chenglou:
@damieng where can we follow on the progress of that?

damieng:
There isn't any visible progress to report just yet but I'll post a link in here once a branch is available.

IanVS:
It's been about a year since the last update, and this feature would be really nice to have.  Nuclide seems to have implemented their own settings package, which includes a filter/search that could perhaps be used as a model.  Is this issue still under consideration, even though it's closed?

damieng:
It is something we want to get to - the real problem is that settings basically needs a rewrite for a whole bunch of reasons.  Search would definitely be a top-billed feature of such a rewrite.

PR #441, opened 2015-03-22
Focus search fields in installed themes and packages panels

### LINKED ISSUE #440: Focus search field after openning install package / theme screen
Hi :),

how about to focusing search field after opening Settings View: install screen? Right now I have to double tab to get there.

mnquintana:
This PR makes it so that the Packages and Themes panels correctly autofocus when clicking on them within settings-view. (#432 will come in another PR.)

(Should I write specs for this?)

### DIFF
--- a/lib/installed-packages-panel.coffee
+++ b/lib/installed-packages-panel.coffee
@@ -53,6 +53,9 @@ class InstalledPackagesPanel extends View
 
     @loadPackages()
 
+  focus: ->
+    @filterEditor.focus()
+
   detached: ->
     @unsubscribe()
 
--- a/lib/themes-panel.coffee
+++ b/lib/themes-panel.coffee
@@ -100,6 +100,9 @@ class ThemesPanel extends View
       @activeUiTheme = @uiMenu.val()
       @scheduleUpdateThemeConfig()
 
+  focus: ->
+    @filterEditor.focus()
+
   detached: ->
     @unsubscribe()
     @disposables.dispose()
--- a/spec/themes-panel-spec.coffee
+++ b/spec/themes-panel-spec.coffee
@@ -6,11 +6,14 @@ Q = require 'q'
 
 PackageManager = require '../lib/package-manager'
 ThemesPanel = require '../lib/themes-panel'
+SettingsView = require '../lib/settings-view'
 
 describe "ThemesPanel", ->
   [panel, packageManager, reloadedHandler] = []
+  settingsView = null
 
   beforeEach ->
+    settingsView = new SettingsView
     atom.packages.loadPackage('atom-light-ui')
     atom.packages.loadPackage('atom-dark-ui')
     atom.packages.loadPackage('atom-light-syntax')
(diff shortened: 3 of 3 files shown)


thomasjo:
> Should I write specs for this?

That would be great as it will insulate us from future regressions. Thanks for fixing this by the way!

mnquintana:
@thomasjo No problem! Although maybe you can help with this - I'm having trouble figuring out how to simulate navigating to a panel. Any ideas?

thomasjo:
There are two easy ways to comes to mind; first is to do e.g.

``` coffee
atom.workspace.open('atom://config/packages')
```

Alternative is to issue a command that triggers basically the same `open` invocation. E.g. `settings-view:uninstall-packages` and similar.

mnquintana:
@thomasjo I don't think that'll work in this case because of #432 (so the test will always fail until #432 is resolved, which I could look into as part of this PR)

@thedaniel Any ideas for simulating navigating to a panel?

thomasjo:
> I don't think that'll work in this case because of #432 

If you don't want to test this in an "end-to-end" fashion, you should be able to leverage [`SettingsView::showPanel`](https://github.com/atom/settings-view/blob/a314e02734ac761c8b86b3af22ba2f506fb0dca0/lib/settings-view.coffee#L166).

thomasjo:
> until #432 is resolved, which I could look into as part of this PR

That would be great, assuming the problem can be fixed in this package.

mnquintana:
@thomasjo @thedaniel Since #432 is a bigger issue than I thought (see https://github.com/atom/settings-view/issues/432#issuecomment-85243483)), I'd much rather work on it in a separate PR. This does resolve #440 though and makes #394 possible, which I'm also working on. What do you think?

thomasjo:
Sounds good to me. I'll let @thedaniel have the final word.

thedaniel:
@mnquintana yep, sounds good.

thedaniel:
To clarify, you want to merge this as it stands now and do further work on top of it in the other PRs right?

mnquintana:
@thedaniel Yup!

thedaniel:
This doesn't seem to fix #440 for me.

```
[settings-view (master)⚡]> git checkout mnquintana/focus-filters 
Note: checking out 'mnquintana/focus-filters'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b new_branch_name

HEAD is now at e4774c1... WIP: Add spec for search filter focus
2015-04-03 14:55:40
[settings-view]> apm link -d
/Users/daniel/.atom/dev/packages/settings-view -> /Users/daniel/atom/settings-view
2015-04-03 14:57:51
[settings-view]> atom -d
```

In the opened window, I run "settings view: install packages and themes" and the search field is not focused.

mnquintana:
@thedaniel You're right, I had been misinterpreting #440 this whole time – it actually is a duplicate of #432. So this resolves an unreported issue, namely, that the Packages and Themes panels didn't autofocus even when clicking on them. I'll update the main PR body to reflect that.

What do you think now?

PR #443, opened 2015-03-25
Settings indicator

### LINKED ISSUE #355: Hide / disable settings button if package / theme has no settings
**Expected:** When I see a Settings button on a package card (like in the Packages or Themes panes' lists), I expect that there will be user-configurable settings for that package.

**Actual:** There may or may not be user-configurable settings for the package – there's no way of knowing without clicking the button (or on the package name link).

I think it would be extremely helpful if those Settings buttons were hidden or disabled if the package doesn't expose any user-configurable settings. You would still be able to easily navigate to the same package detail view by clicking on the package name link in the package card. This would be a simple visual indicator for packages without user-configurable settings, and would keep users from wasting too much time clicking through looking for configurable settings.

#### Hidden
image

#### Disabled
image

mnquintana:
Resolves #355 

image

I kept the style of the settings button as-is, but we could remove the text if that would be better. Personally, I think we should keep it for clarity. (Maybe remove it on smaller viewport widths?)

I could also add the "scroll to settings" behavior to this PR as well - let me know if you think that would be best.

/cc @thedaniel @simurai @zephraph

### DIFF
--- a/lib/package-card.coffee
+++ b/lib/package-card.coffee
@@ -68,7 +68,10 @@ class PackageCard extends View
     if atom.packages.isBundledPackage(@pack.name) and @type is 'theme'
       @statusIndicator.hide()
 
-    if opts?.onSettingsView or not @hasSettings(@pack)
+    unless @hasSettings(@pack)
+      @settingsButton.hide()
+
+    if opts?.onSettingsView
       @settingsButton.remove()
     else
       @on 'click', =>
--- a/lib/package-card.coffee
+++ b/lib/package-card.coffee
@@ -37,8 +37,8 @@ class PackageCard extends View
           @div class: 'btn-group', =>
             @button type: 'button', class: 'btn btn-info icon icon-cloud-download install-button', outlet: 'installButton', 'Install'
           @div outlet: 'buttons', class: 'btn-group', =>
-            @button type: 'button', class: 'btn icon icon-gear settings',           outlet: 'settingsButton', 'Settings'
-            @button type: 'button', class: 'btn icon icon-trashcan uninstall',       outlet: 'uninstallButton', 'Uninstall'
+            @button type: 'button', class: 'btn icon icon-gear settings',             outlet: 'settingsButton', 'Settings'
+            @button type: 'button', class: 'btn icon icon-trashcan uninstall',        outlet: 'uninstallButton', 'Uninstall'
             @button type: 'button', class: 'btn icon icon-playback-pause enablement', outlet: 'enablementButton', =>
               @span class: 'disable-text', 'Disable'
             @button type: 'button', class: 'btn status-indicator', tabindex: -1, outlet: 'statusIndicator'
--- a/lib/package-card.coffee
+++ b/lib/package-card.coffee
@@ -191,7 +191,7 @@ class PackageCard extends View
   isDisabled: -> atom.packages.isPackageDisabled(@pack.name)
 
   hasSettings: (pack) ->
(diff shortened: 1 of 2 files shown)


zephraph:
Nice! :+1:

simurai:
> I could also add the "scroll to settings" behavior to this PR as well

I think it's a good idea. Is it hard to implement/test? But could also be part of a different PR.

mnquintana:
> I think it's a good idea. Is it hard to implement/test? But could also be part of a different PR.

I wasn't quite sure how to implement it, since it isn't *really* loading a URI on click so it wouldn't be as simple as tacking an element ID on the end. I was actually strongly considering using this semi-experimental DOM method: http://devdocs.io/dom/element.scrollintoview which would fit the bill nicely.

mnquintana:
I'm thinking now that I'm gonna add the "scroll to settings behavior" in another PR – I don't really want to block this enhancement from being merged.

So this is ready for review!

jlord:
I think it would still be useful to get to the packages readme, even if there aren't settings per say. You could still get it it (I'm assuming) through clicking the body of the package card, but that isn't very straight forward. Though I also agree a 'Settings' button isn't very straightforward when really there's just a readme. 

What about replacing the Settings button with a "Readme" button in these cases?

mnquintana:
@jlord I think that would be a little confusing – then it would seem that packages with Settings buttons don't have Readmes, that Settings and Readmes are somehow mutually exclusive. You can still get to the Readme pretty easily from clicking the package card body – I'd expect that to be the primary way of getting to the Readme to learn more about a package (it really is quite intuitive especially in the Install panel).

It might be worth looking into adding an additional Readme indicator to the package card, but I think that would be a bit out of scope for this PR, and now that I think about it, I've never seen a package without a Readme so it might not add that much value?

mnquintana:
Actually, I think there used to be a Readme button under the package card at the top of the package detail view – we could bring that button back and have it scroll to the Readme?

jlord:
> it would seem that packages with Settings buttons don't have Readmes

Yeah, agree, not ideal. 

I just pulled the branch down and think that clicking on the card will be more natural than I originally thought. And :+1: to pass on the readme bit for this PR.

mnquintana:
@jlord Awesome! Do you think this is good to merge?

lib/package-card.coffee
```diff
@@ -187,6 +190,9 @@ class PackageCard extends View
 
   isDisabled: -> atom.packages.isPackageDisabled(@pack.name)
 
+  hasSettings: (pack) ->
+    if atom.config.get(pack.name)? then true else false
```
kevinsawicki: Minor style note here, this line could just be:

```coffee
hasSettings: (pack) ->
    atom.config.get(pack.name)?
```

Ending with a `?` check guarantees a boolean will always be returned.

lib/package-card.coffee
```diff
@@ -68,6 +68,9 @@ class PackageCard extends View
     if atom.packages.isBundledPackage(@pack.name) and @type is 'theme'
       @statusIndicator.hide()
 
+    unless @hasSettings(@pack)
+      @settingsButton.hide()
```
jlord: was just talking with @simurai and let's use `.remove()` here because it will help the button styles round corners correctly when one button is left.

mnquintana:
@jlord @kevinsawicki Should be ready for review again!

jlord:
Looks good, thanks @mnquintana! :balloon:

PR #450, opened 2015-04-02
Use .remove() rather than .hide() for sake of button group

jlord:
Instead of _hiding_ certain buttons, we should _remove_ them so that they do not interfere with how the border radius is applied to the button group.

Before & After

screen shot 2015-04-01 at 12 44 38 pm

screen shot 2015-04-02 at 1 50 24 pm

This:
- `.remove()` `install` and `uninstall` buttons from bundled packages.
- `.remove()` `statusIndicator` and `enablement` from themes since you cannot toggle them enabled/disabled.
- adds and updates specs to verify themes do not get `settings` `enablement` or `statusIndicator` buttons

### DIFF
--- a/spec/available-package-view-spec.coffee
+++ b/spec/available-package-view-spec.coffee
@@ -14,7 +14,17 @@ describe "PackageCard", ->
   it "doesn't show the disable control for a theme", ->
     setPackageStatusSpies {installed: true, disabled: false}
     view = new PackageCard {theme: 'syntax', name: 'test-theme'}, @packageManager
-    expect(view.enablementButton.css('display')).toBe('none')
+    expect(view.find.enablementButton).not.toExist()
+
+  it "doesn't show the status indicator for a theme", ->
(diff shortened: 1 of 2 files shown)


jlord:
cc @simurai @thedaniel

spec/available-package-view-spec.coffee
```diff
@@ -42,7 +52,7 @@ describe "PackageCard", ->
     view.installButton.click()
     expect(@packageManager.install).toHaveBeenCalled()
 
-  it "hides the settings button if a package has no settings", ->
+  it "removes the settings button if a package has no settings", ->
     setPackageStatusSpies {installed: true, disabled: false, hasSettings: false}
     view = new PackageCard {name: 'test-package'}, @packageManager
     expect(view.find('.btn.settings')).not.toExist()
```
mnquintana: Whoops, I didn't realize you could do `view.find.settingsButton` so I did it the hard way here – we should probably change this.

PR #458, opened 2015-04-04
Hiding of settings with unfulfilled dependencies.

leipert:
As the package config is written in JSON Schema, I propose the addition of [Property Dependencies](http://json-schema.org/latest/json-schema-validation.html#anchor70).

Imagine you have the following config. You just want to show `options` when `enabled` and `alsoEnabled` are both true. And you want to show `options.path` when `options.enabled` is true.

```coffee
module.exports =
  config:
    enabled:
      type: 'boolean'
      default: false
      order: 1
    alsoEnabled:
      type: 'boolean'
      default: false
      order: 2
    options:
      dependencies: ['enabled', 'alsoEnabled']
      type: 'object'
      order: 3
      properties:
        enabled:
          type: 'boolean'
          default: false
        path:
          dependencies: ['enabled']
          type: 'string'
          default: '/random/path'
```

In Action:

Probably I did something wrong and there is surely room for improvement. But I'd like to get this discussion started, as it would un-clutter some of my packages.

### DIFF
--- a/lib/settings-panel.coffee
+++ b/lib/settings-panel.coffee
@@ -36,6 +36,7 @@ class SettingsPanel extends View
     @bindInputFields()
     @bindSelectFields()
     @bindEditors()
+    @bindDependencies()
 
   detached: ->
     @disposables.dispose()
@@ -169,6 +170,18 @@ class SettingsPanel extends View
 
     value
 
+  bindDependencies: (namespace, settings) ->
+    @find('.control-group').toArray().forEach (group) =>
+      group = $(group)
+      keynames = group.attr('class').replace(/\s*control-group\s*/ig, '')
+      if keynames.length > 0
+        keynames = keynames.split(" ")
+        keynames.forEach (keyname) ->
+          atom.config.observe keyname, (value) ->
+            show = if keynames.length is 1 then !!value else checkAllDependencies(keynames)
+            if show then group.show() else group.hide()
+
+
 ###
 # Space Pen Helpers
 ###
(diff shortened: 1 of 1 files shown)


thedaniel:
This concept (nested config) is appealing to me, though I hesitate to add any more complexity to the sea of checkboxes that is the options pane. An alternative might be disabling dependent controls rather than hiding them - if there's any kind of complex dependency relationship controls could pop in and out of the view in ways that might be confusing to the end user.

Or, if we pursue this, maybe we can simplify the API by starting with only allowing single boolean dependencies and seeing how much use it gets?

maxbrunsfeld:
> An alternative might be disabling dependent controls rather than hiding them.

:+1: This seems like a better UI than controls showing and hiding as you interact with other controls.

thedaniel:
@atom/feedback how do you feel about this? I'm in favor of it, because I think it will help make settings less confusing in situations like this, with these changes:

- Show unavailable dependent controls as disabled rather than hiding them (as discussed above)
- rename `depedencies` to `dependsOn` or `requires` since that's the direction the dependency goes
- limit this behavior to boolean options

If a quorum of folks think this is a good idea, I'm happy to polish it and add specs / guide @leipert through that, since it seems like a small change that leads to a UX improvement with a non-breaking and easy to understand config schema change.

leipert:
I really like that there is movement on this :+1:.

Your proposals look good to me, @thedaniel, but I would like to stay with the `dependencies` keyword, so that the settings stay valid json schema in the future. At least the atom docs [mention](https://atom.io/docs/api/v1.0.19/Config#config-schemas), that it is json schema.

I am happy to adjust this PR and write specs/documentation changes, just tell me what to do :)

maxbrunsfeld:
@leipert If we want to remain compatible with `json-schema`, which I think we should, the `dependencies` key needs to work differently. [Here's](https://github.com/json-schema/json-schema/wiki/Dependencies#property-dependencies) one example I found.

leipert:
That is true. Maybe `requires` or something like that is better.
It would also be better to expose the full settings key for transparency reasons. So instead some random scoped `enabled` matching I did in the initial PR, more something like 

Maybe something along the lines of:

```coffeescript
module.exports =
  config:
    enabled:
      type: 'boolean'
      default: false
      order: 1
    alsoEnabled:
      type: 'boolean'
      default: false
      order: 2
    options:
       # depends on enabled and alsoEnabled from the root config
      requires: ['enabled', 'alsoEnabled']
      type: 'object'
      order: 3
      properties:
        enabled:
          type: 'boolean'
          default: false
        path:
          # Depends on options.enabled and not the enabled from root config
          requires: 'options.enabled' 
          type: 'string'
          default: '/random/path'
```

leipert:
aaah, `dependsOn` is probably better than `requires` to avoid confusion with `required`

damieng:
This feature seems quite interesting - is it still alive or is this WIP PR abandoned?

leipert:
It's abandoned from my side. I do not fiddle around with atom too much anymore. Feel free to take over on this PR.

Issue #465, opened 2015-04-14
Uncaught TypeError: undefined is not a function


mnquintana:
*(Note: Not a duplicate of #359)*

[Enter steps to reproduce below:]

1. Open Settings > Install 
2. Find package that you haven't installed yet
3. Click on package card body

I have also gotten this in safe mode.

**Atom Version**: 0.191.0-a68e837
**System**: Mac OS X 10.10.3
**Thrown From**: [settings-view](https://github.com/atom/settings-view) package, v0.191.0

### Stack Trace

Uncaught TypeError: undefined is not a function

```
At /Applications/Atom.app/Contents/Resources/app/node_modules/settings-view/lib/package-detail-view.js:155

TypeError: undefined is not a function
  at PackageDetailView.module.exports.PackageDetailView.activate (/Applications/Atom.app/Contents/Resources/app/node_modules/settings-view/lib/package-detail-view.js:155:26)
  at PackageDetailView.module.exports.PackageDetailView.initialize (/Applications/Atom.app/Contents/Resources/app/node_modules/settings-view/lib/package-detail-view.js:145:12)
  at PackageDetailView.View (/Applications/Atom.app/Contents/Resources/app/node_modules/atom-space-pen-views/node_modules/space-pen/lib/space-pen.js:184:25)
  at new PackageDetailView (/Applications/Atom.app/Contents/Resources/app/node_modules/settings-view/lib/package-detail-view.js:38:54)
  at /Applications/Atom.app/Contents/Resources/app/node_modules/settings-view/lib/settings-view.js:282:22
  at SettingsView.module.exports.SettingsView.getOrCreatePanel (/Applications/Atom.app/Contents/Resources/app/node_modules/settings-view/lib/settings-view.js:287:19)
  at SettingsView.module.exports.SettingsView.showPanel (/Applications/Atom.app/Contents/Resources/app/node_modules/settings-view/lib/settings-view.js:323:24)
  at space-pen-div.<anonymous> (/Applications/Atom.app/Contents/Resources/app/node_modules/settings-view/lib/package-card.js:175:26)
  at space-pen-div.jQuery.event.dispatch (/Applications/Atom.app/Contents/Resources/app/node_modules/jquery/dist/jquery.js:4430:9)
  at space-pen-div.elemData.handle (/Applications/Atom.app/Contents/Resources/app/node_modules/jquery/dist/jquery.js:4116:28)

```

### Commands

```
[...]

PR #466, opened 2015-04-14
Check if package is loaded before activating config

izuzak:
Fixes https://github.com/atom/settings-view/issues/465.

Travis is refusing to run the tests for an unknown reason. However, I ran them locally and I'm getting failures which are also present on master (they don't seem to be related to this PR).

cc @nathansobo @mnquintana


thedaniel:
wow, thanks, what a bug :/

PR #468, opened 2015-04-15
Make settings panels scrollable with keyboard.

brettle:
Fixes https://github.com/atom/settings-view/issues/544


kankaristo:
Does this add the `home` and `end` keys to scroll to the top and bottom of the view?

I constantly find myself pressing them (mostly `home`) to move around, and press them a few times before once again realizing that they don't actually do anything.

brettle:
PgUp and PgDn work consistently, but whether and when Home and End work depends on which panel you are looking at. They work immediately in the panel called "Settings", but they don't work in the other standard panels (Keybindings, Packages, Themes, Updates, Install) unless you click in the panel first. In most of those cases, this is because those panels focus a textbox when the panel comes up. As a result, Home and End go to the textbox until you click somewhere else.

kankaristo:
I can only get Home and End working in the Keybindings panel, not in any of the other panels (even after clicking in them). In Keybindings, I have to click on the list of key bindings. Clicking anywhere else (e.g. next to the "Keybindings" header) doesn't work.

In a package's settings panel Home and End start working if I click out of the automatically focused text box. I hadn't noticed that before, thanks for bringing it up! :)

izuzak:
@brettle It seems that this PR no longer merges without conflicts. Can you merge master into your branch and resolve those conflicts so that someone can review the changes?

brettle:
@izuzak Should be fixed now.

maxbrunsfeld:
For me, this PR does not change the scrolling behavior that I see on master. As @kankaristo said,

> I can only get Home and End working in the Keybindings panel, not in any of the other panels (even after clicking in them). In Keybindings, I have to click on the list of key bindings. Clicking anywhere else (e.g. next to the "Keybindings" header) doesn't work.

brettle:
@maxbrunsfeld Sorry for the delay. This PR no longer seems to change the scrolling behavior. I'm guessing something changed in the core to make it stop working, but I haven't had a chance to look into it yet.

brettle:
After further investigation it appears that commit d8c041a7d57ae942600218e039461763bc79771d by @jlord (courtesy @simurai?) caused this PR to stop working. That commit moved the "overflow: auto" property from the ".panels" div that contains all of the settings panels, to the child divs (".panels > div") that display each settings panel. My PR makes the parent div a ScrollView so that it responds to keyboard scroll events. However, if the parent div is not scrollable (e.g. via overflow:auto) this has no effect. I'm not sure what motivated the aforementioned commit, so I'm not sure whether my PR should revert it. If it shouldn't be reverted, an alternative is to change each of the individual settings panels (e.g. GeneralPanel, KeybindingsPanel, etc) to extend ScrollView instead of View. I'm happy to do that but you'd also need to remember do that for any future settings panels.

Thoughts?

jlord:
@brettle That commit came from #490 and the parent div not actually containing its children. @simurai might have better thoughts than me on having both things work (not breaking the fix that #490 provides and enabling keyboard scrolling).

simurai:
If I remember correctly, #490 fixes an issue where it wasn't always scrolled to the top when switching between package settings. The fix was to move `overflow: auto;` from `.panels` to the child `<div>`.

@brettle Would keyborad scrolling work if the child `<div>` would become a ScrollView?

Using `.panels > div` feels a bit too vague and those divs should probably get their own class name. How about `.panels-item` since it seems to be a wrapper > items relationship?

So instead of

```html
<div class="panels">
  <div is="space-pen-div">
  <div is="space-pen-div">
</div>
```

add

```html
<div class="panels">
  <div class="panels-item" is="space-pen-div">
  <div class="panels-item" is="space-pen-div">
</div>
```

brettle:
Thanks @simurai and @jlord. I've updated the PR to move the keyboard scrolling into the package panels (i.e. the child divs become ScrollViews) and added the .panels-item CSS class that @simurai suggested. Manual and automated tests pass.

simurai:
Tried it out and keyboard navigation works. :+1: 

Something that doesn't work for me is clicking on the package cards. It just scrolls to the top. Also, when clicking outside of the cards:

settings-1

Clicking on the "Settings" button works, but the package (detail) settings can't be scrolled. I think it's because it misses the `panels-item` class that adds overflow scrolling. Probably [here](https://github.com/brettle/settings-view/blob/master/lib/package-detail-view.coffee#L24).

screen shot 2015-07-16 at 10 07 22 pm

brettle:
Thanks for the feedback @simurai. The problems you mentioned have now been fixed. It took me forever to figure out what was causing the "scrolling to the top" problem. See [the comment in my latest commit](https://github.com/brettle/settings-view/blob/dba944cd5db6ad5de61454e9b62bbd0ad8fb63ac/lib/settings-view.coffee#L29) if you are curious.

simurai:
:+1: 

Is it possible to make `home` and `end` work for all views? Would be handy to quickly reach the bottom of packages to see the Development Packages. Or only if there is no auto focus on the search?

brettle:
Ctrl-Home and Ctrl-End work on all views. Home and End will only work on all views if there is no auto focus on the search.

izuzak:
@simurai wondering how you feel about this -- anything else you think should be done? :thought_balloon:

simurai:
Hmm.. hitting `end` in package/themes would've been nice so you can reach the dev packages very quickly. But removing the auto-focus on the search will probably not get any votes?

Otherwise I think we should merge it, since it's already a big improvement. Anyone wants to do a last code check?

thomasjo:
> Anyone wants to do a last code check?

Looks good to me!

wonderbeyond:
I want it!

thedaniel:
Sorry to leave this so long. I see 22 failures for this on Atom master. I'll fork and take a look.

thedaniel:
Oh sorry, they are green after merging master in again.

lib/settings-view.coffee
```diff
@@ -26,7 +26,14 @@ class SettingsView extends ScrollView
...
           @button class: 'btn btn-default icon icon-link-external', outlet: 'openDotAtom', 'Open Config Folder'
-      @div class: 'panels', outlet: 'panels'
+      # The tabindex attr below ensures that clicks in a panel item won't cause this view to gain focus.
+      # This is important because when this view gains focus (e.g. immediately after atom displays it),
+      # it focuses the currently active panel item. If that focusing causes the active panel to scroll (e.g.
+      # because the active panel itself passes focus on to a search box at the top of a scrolled panel),
+      # then the browser will not fire the click event on the element within the panel on which the user originally
+      # clicked (e.g. a package card). This would prevent us from showing a package detail view when clicking on a
+      # package card. Phew!
+      @div class: 'panels', tabindex: -1, outlet: 'panels'
```
thedaniel: Great, useful comment. This view in general could use more of this ;)

thedaniel:
Finished up and merged in #677. Thank you, @brettle and thanks for your patience with the long-running review!

brettle:
Your welcome. Keep up the great work on Atom!

PR #470, opened 2015-04-16
Normalize cursor and select for headings and package cards

mnquintana:
This disables select for package cards and headings, and uses the default cursor for package cards in the detail view.

/cc @simurai @thedaniel


simurai:
:+1:

PR #473, opened 2015-04-22
Use native keybindings in Readme preview

### LINKED ISSUE #469: Can't copy text from Readme preview
Just noticed this today when I was showing Atom to a colleague – there's currently no way to copy and paste text from the Readme preview for a package.

mnquintana:
Fixes #469 

Also removes an extraneous `div`.


thedaniel:
looks good to me!

PR #474, opened 2015-04-22
Fix config menu user select

mnquintana:
The config menu was actually user selectable before (since it was using the un-prefixed property).

/cc @simurai @thedaniel

### DIFF
--- a/styles/settings-view.less
+++ b/styles/settings-view.less
@@ -310,6 +310,7 @@
     flex: 1.6;
     min-width: 180px;
     max-width: 240px;
+    -webkit-user-select: none;
     user-select: none;
     padding-top: @component-padding;
     background-color: lighten(@base-background-color, 2%);


simurai:
I wonder if we shouldn't just do:

```css
html {
  -webkit-user-select: none;
  cursor: default;
}
```

Make everything un-selectable by default, and then enable selection in certain areas. Like the notifications and panes. Or is that too risky.

mnquintana:
@simurai Hmmm that might be a better way to go, but that would be a much bigger change in core. Although FWIW, [that is apparently what Brackets does.](https://github.com/adobe/brackets/blob/master/src/styles/brackets.less#L52)

/cc @kevinsawicki @nathansobo @maxbrunsfeld @benogle What do you think?

thedaniel:
I am going to merge this because it fixes the immediate bug without changing the user-select strategy wholesale, and we can have the `html {}` discussion elsewhere.

PR #490, opened 2015-04-30
Ensure package-detail contains its children

### LINKED ISSUE #472: Clicking on a package doesn't reset the scrollbar
Expected: Clicking on a package will reset you to the top of whatever package you clicked on
Actual: Clicking on a package will put you wherever your scrollbar was while clicking on it...kinda.

Here's a GIF (with some issues listed below):
1) The GIF started a bit late so instead of seeing me scroll down to markdown-preview it starts in the middle of me scrolling back up to Packages
2) For some reason my mouse didn't show up
3) My computer froze for a second when clicking on language-clojure so ignore that period where the GIF doesn't do anything
settings-scroll

Atom 0.193.0-c795879, safe mode, Ubuntu 14.10.

jlord:
Came across the need to add this (the div was ending before all of its children) while investigating the _click `package-card` and go to detail view only to have your scroll bar somewhere in the middle_ issue. 

Adding this seems to take care of the scrolling issue and you now start at the top of the detail view. 

_package-detail, before fix, ending too early_

**To Reproduce**
- Open Settings 
- Click Packages, scroll to very last package in list and click the package card
- See that you're in the middle of that package's detail/settings view.

Repeat on this branch, find that you start at the top of the detail/settings view as expected.

cc @simurai for CSS guidance—do you know why this helps and if I've used it correctly?

Closes #472

### DIFF
--- a/styles/available-package.less
+++ b/styles/available-package.less
@@ -4,6 +4,10 @@
 
 .settings-view {
 
+  .package-detail {
+    overflow-y: auto;
+  }
+
   // TODO: Somehow unify the "card".
 
   .package-card {
--- a/styles/available-package.less
+++ b/styles/available-package.less
@@ -4,10 +4,6 @@
 
 .settings-view {
 
-  .package-detail {
-    overflow-y: auto;
-  }
-
   // TODO: Somehow unify the "card".
 
   .package-card {
--- a/styles/settings-view.less
+++ b/styles/settings-view.less
@@ -403,11 +403,11 @@
     flex-flow: column;
     display: flex;
     position: relative;
-    overflow: auto;
 
     > div {
       flex: 1;
       min-width: 372px; // magic number: fit the Settings, Uninstall and Disable button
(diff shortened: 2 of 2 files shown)


simurai:
Yeah, that seems to fix it. :+1: 

Something I'm not sure.. it would add scrolling inside a scrolling area. Haven't seen it, but that might cause issues.. like double scrollbars or so.

Another thing that seems to fix it is to move the scrolling from the parent `.panels` to the `div`s.

```less
// Line 414 in settings-view.less
.panels {
    flex: 8.4;
    flex-flow: column;
    display: flex;
    position: relative;
    // move from here ->

    > div {
      flex: 1;
      min-width: 372px;
      overflow: auto; // <- to here
    }
  }
```

I think there is never a situation where more than one `div` is visible, so having the scrolling inside it, should be fine.

jlord:
@simurai :sparkles:

PR #502, opened 2015-05-09
Prevent repo link being full width

simurai:
Before:

minimap-before

After:

minimap-after


lee-dohm:
:+1: Looks good to me

PR #508, opened 2015-05-19
Upgrade Bootstrap

simurai:
Fixes checkbox position. Depends on https://github.com/atom/atom/pull/6848

Tested on several themes. All seem good except in Isotope the `.setting-description` is mis-aligned.


thomasjo:
LGTM!

PR #509, opened 2015-05-19
Add upgrade notification

benogle:
This will show when atom starts up if there are packages to upgrade

screen shot 2015-05-19 at 3 23 59 pm


mnquintana:
:joy: :+1:

Issue #505, opened 2015-05-15
Git repository link trimmed


jerone:
The [repository link](https://github.com/atom/settings-view/blob/9b8bfe1b5727b58f1fc2a558d94e24c3fe99c289/lib/package-manager.coffee#L306) is incorrectly [trimmed](https://github.com/atom/settings-view/blob/master/lib/package-detail-view.coffee#L90) at the beginning of the link when it's a `git@github.com:` URI. 
Example: [`git@github.com:jerone/menu-manager.git`](https://github.com/jerone/menu-manager/blob/386b01aad8c85c49ba7153f44d43dba08229b9b3/package.json#L21)

![Git repository link trimmed](https://cloud.githubusercontent.com/assets/55841/7652182/085be1ba-fb0a-11e4-8653-488d64802131.png)

Windows 7 64-bits
Atom 0.198.0

jlord:
Strange... @mnquintana maybe, but not sure why just the 'g' would be getting trimmed. I'll dig into it today.

jerone:
[**@jlord**](https://github.com/jlord) commented on [15 mei 2015 18:01 CEST](https://github.com/atom/settings-view/issues/505#issuecomment-102448278 "2015-05-15T16:01:52Z - Replied by Github Comment Enhancer"):
> Strange... [@mnquintana](https://github.com/mnquintana) maybe, but not sure why just the 'g' would be getting trimmed. I'll dig into it today.

The `g` is trimmed because of the code is only accepting `http(s)://` urls, which then is stripped to a [pathname](https://github.com/atom/settings-view/blob/master/lib/package-detail-view.coffee#L89) and the leading slash is [trimmed](https://github.com/atom/settings-view/blob/master/lib/package-detail-view.coffee#L90).

PR #512, opened 2015-05-20
Handle Git Repository URLs

jlord:
This handles Git urls (i.e., `git@github.com:atom/settings-view`) used in the `repository` field for packages. 

Fixes https://github.com/atom/settings-view/issues/505

Links work in `package-detail-view` so I think these two spots should cover it. Before the avatar and repo link were broken. 

screen shot 2015-05-20 at 12 08 43 pm

It may be a bit verbose and handling the one off case like this may not be ideal to here if there are like 4 other types of urls that may also come through. Thoughts?

cc @kevinsawicki @thedaniel

### DIFF
--- a/lib/package-card.coffee
+++ b/lib/package-card.coffee
@@ -149,6 +149,9 @@ class PackageCard extends View
       repo = repository
     else
       repo = repository.url
+      if repo.match 'git@github'
+        repoName = repo.split(':')[1]
+        repo = "https://github.com/#{repoName}"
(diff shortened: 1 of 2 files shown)


thedaniel:
> handling the one off case like this may not be ideal to here if there are like 4 other types of urls that may also come through.

IMO It works, it's easy to read, it's fine - if people see breakage with other url types we can rethink it later but this seems to cover the 99.9% case.

jlord:
:+1:

PR #526, opened 2015-06-04
Fix avatarGlob to not pick up cached avatars with prefix collisions

### LINKED ISSUE #482: Core packages sporadically use community package avatar if org name contains "atom"
This is sort of a strange one, but I've seen it for quite a long time:

screenshot_2015-04-26_17_53_10

The avatar icon for toggle-quotes is actually [atom-minimap](https://github.com/atom-minimap)'s avatar. I've also seen this happen with AtomLinter's and atom-community's avatars too. This only happens if the package's org name happens to contain (or maybe start with) `atom`.

Possible steps to reproduce:
1. Ensure that minimap is installed.
2. Look at avatar icon for a core package.

/cc @thedaniel

jbt:
This fixes #482.

The glob pattern of `#{login}-*` was erroneously picking up cache files named (for example) `atom-minimap-1234567890` when looking for a file matching `atom-{TIMESTAMP}` (or anyone whose login happened to start with `atom-`).

Changing it to only match trailing digits means the wildcard will only match the timestamp, not timestamp plus any other chars if `#{login}-` happens to be a prefix of somebody else's username.

### DIFF
--- a/lib/atom-io-client.coffee
+++ b/lib/atom-io-client.coffee
@@ -128,7 +128,7 @@ class AtomIoClient
       callback(null, null)
 
   avatarGlob: (login) ->
-    path.join @getCachePath(), "#{login}-*"
+    path.join @getCachePath(), "#{login}-*([0-9])"
 
   fetchAndCacheAvatar: (login, callback) ->
     if @online()


mnquintana:
I was wondering how we might write a spec for something like this... would we have to diff the image binaries? That might not be worth it :grin:

maxbrunsfeld:
@mnquintana Yeah, I don't think we need a spec for this one.

maxbrunsfeld:
@jbt Thanks for figuring this out!

PR #538, opened 2015-06-09
Replace emissary with event-kit

### LINKED ISSUE #537: Migrate from emissary to event-kit
[atom/emissary](https://github.com/atom/emissary) is being deprecated in favour of [atom/event-kit](https://github.com/atom/event-kit), so `settings-view` should be updated to use that.

mgarciaisaia:
Replaces the deprecated `emissary` lib with `event-kit`.

The first one is still a dependency as it's needed in `install-panel.coffee` for [subscribing to a third-party](https://github.com/mgarciaisaia/settings-view/blob/537-event-kit/lib/install-panel.coffee#L76-L77) - or isn't it?

Fixes #537

### DIFF
--- a/lib/install-panel.coffee
+++ b/lib/install-panel.coffee
@@ -3,7 +3,7 @@ path = require 'path'
 _ = require 'underscore-plus'
 fs = require 'fs-plus'
 {$, $$, TextEditorView, View} = require 'atom-space-pen-views'
-{Subscriber} = require 'emissary'
+{CompositeDisposable} = require 'atom'
 
 PackageCard = require './package-card'
 Client = require './atom-io-client'
@@ -14,7 +14,6 @@ PackageNameRegex = /config\/install\/(package|theme):([a-z0-9-_]+)/i
 
 module.exports =
 class InstallPanel extends View
-  Subscriber.includeInto(this)
 
   @content: ->
     @div =>
@@ -47,6 +46,7 @@ class InstallPanel extends View
           @div outlet: 'featuredContainer', class: 'container package-container'
 
   initialize: (@packageManager) ->
+    @disposables = new CompositeDisposable()
     client = $('.settings-view').view()?.client
     @client = @packageManager.getClient()
     @atomIoURL = 'https://atom.io/packages'
@@ -63,17 +63,16 @@ class InstallPanel extends View
     @loadFeaturedPackages()
 
   detached: ->
-    @packageManagerSubscription.dispose()
-    @unsubscribe()
+    @disposables.dispose()
 
   focus: ->
     @searchEditorView.focus()
 
   handleSearchEvents: ->
-    @packageManagerSubscription = @packageManager.on 'package-install-failed', (pack, error) =>
+    @disposables.add @packageManager.on 'package-install-failed', (pack, error) =>
       @searchErrors.append(new ErrorView(@packageManager, error))
 
-    @subscribe atom.commands.add @searchEditorView.element, 'core:confirm', =>
+    @disposables.add atom.commands.add @searchEditorView.element, 'core:confirm', =>
       @performSearch()
 
     @searchPackagesButton.on 'click', =>
--- a/lib/install-panel.coffee
+++ b/lib/install-panel.coffee
@@ -63,13 +63,14 @@ class InstallPanel extends View
     @loadFeaturedPackages()
 
   detached: ->
+    @packageManagerSubscription.dispose()
     @unsubscribe()
 
   focus: ->
     @searchEditorView.focus()
 
   handleSearchEvents: ->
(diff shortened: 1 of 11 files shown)


lib/installed-packages-panel.coffee
```diff
@@ -1,6 +1,6 @@
 _ = require 'underscore-plus'
 {$$, TextEditorView, View} = require 'atom-space-pen-views'
-{Subscriber} = require 'emissary'
+{CompositeDisposable} = require 'event-kit'
```
mnquintana: Would you mind requiring this from `'atom'` instead and remove the explicit `event-kit` dependency?

mgarciaisaia: Sure!

Just followed the libraries' README example. May we need to add a comment
about this in there?

I'll update this later - thanks!

lib/package-detail-view.coffee
```diff
@@ -120,16 +118,16 @@ class PackageDetailView extends View
     @sections.append(new PackageReadmeView(readme))
 
   subscribeToPackageManager: ->
-    @subscribe @packageManager, 'theme-installed package-installed', (pack) =>
+    @packageManager.on 'theme-installed package-installed', (pack) =>
```
maxbrunsfeld: Why did you not save off the `Disposable` returned by `.on()` in this case?

lib/package-manager.coffee
```diff
@@ -384,7 +384,13 @@ class PackageManager
...
     theme = pack.theme ? pack.metadata?.theme
     eventName = if theme then "theme-#{eventName}" else "package-#{eventName}"
-    @emit eventName, pack, error
+    @emitter.emit eventName, pack, error
+
+  on: (selectors, callback) ->
+    subscriptions = new CompositeDisposable
+    selectors.split(" ").forEach (selector) =>
+      subscriptions.add @emitter.on selector, callback
+    subscriptions
```
maxbrunsfeld: Normally, for objects that use `event-kit` and emit events, we define event-specific methods for subscribing, like this:

```coffee
onDidInstallAlternativePackage: (callback) ->
  @emitter.on('package-installed-alternative', callback)
```

We would just need to change callers of `PackageManager` to use this method, rather than `.on()`. Would you mind tweaking the style in this way?

lib/install-panel.coffee
```diff
@@ -63,13 +63,14 @@ class InstallPanel extends View
...
 
   focus: ->
     @searchEditorView.focus()
 
   handleSearchEvents: ->
-    @subscribe @packageManager, 'package-install-failed', (pack, error) =>
+    @packageManagerSubscription = @packageManager.on 'package-install-failed', (pack, error) =>
       @searchErrors.append(new ErrorView(@packageManager, error))
 
     @subscribe atom.commands.add @searchEditorView.element, 'core:confirm', =>
```
maxbrunsfeld: You don't need to keep the `Subscriber` mix-in for this. You can use `event-kit` here too:

```coffee
@disposables.add atom.commands.add @searchEditorView.element, 'core:confirm', =>
  @performSearch()
```

You just need to this in the `initialize` method:

```coffee
@disposables = new CompositeDisposable
```

This way, you can get rid of the dependency on `emissary`.

maxbrunsfeld:
@mgarciaisaia Thanks for doing this! Left some minor comments, but other than those, this looks good.

- @maxbrunsfeld and @jlord

lib/package-detail-view.coffee
mgarciaisaia: I was drunk, I hope.

Just forgot about that - I event commented out the `@unsubscribe()` part and commited that :fearful:

lib/install-panel.coffee
mgarciaisaia: So both libs are interchangeable? I thought it would be needed to first migrate the `@searchEditorView` - that is not provided by `settings-view` - to `event-kit` before replacing this.

lib/package-manager.coffee
mgarciaisaia: I fear things [like this](https://github.com/atom/settings-view/blob/37aad1d2b84b95ca83e5f9b43f10fd282ecf9fcb/lib/package-card.coffee#L351-L371) would become _really_ difficult to implement.

I'm going to change this to use `onXXX` methods, but it really smells to me :/

maxbrunsfeld: I see what you're saying. Don't worry about changing this actually.

mgarciaisaia:
@maxbrunsfeld I think I've pretty much addressed everything - may we open a new issue to discuss how to handle the `onXXX` subscriptions?

Thanks for the feedback :)

benogle:
Thanks these commits have been merged

mgarciaisaia:
:+1: great :)

PR #539, opened 2015-06-10
Settings button works on Available Updates

### LINKED ISSUE #499: Settings button in update view
..does not work after having updated the package. For example, I just updated One-Dark-UI but the settings button does nothing in that view now.

mgarciaisaia:
Both the Settings button and the panel itself can be clicked to access the package detail panel.

Fixes #499


mgarciaisaia:
Replaced by #540

PR #540, opened 2015-06-10
Use PackageCard in Available Updates panel

### LINKED ISSUE #499: Settings button in update view

### LINKED ISSUE #529: Uninstall button doesn't work in package list
Woks only in package settings view("Settings" button).

Also cannot open package settings in "Updates" tab.

win 10; atom 0.207

mgarciaisaia:
PackageCards show Update buttons whenever there are new versions (even in other panels).

update-cards

They cards aren't half-wided, but they work better than the previous view.

Fixes #499
Fixes #529
Supersedes #539


mnquintana:
Awesome, thanks so much for this! I did find a few issues though:

- Package card doesn't change state when clicking the uninstall button
  - With disabled packages, the package card doesn't change state at all
  - Uninstalling enabled packages leaves a package card with an update button
    - Clicking that update button re-installs the package and updates it

mgarciaisaia:
I sure can try.

May those be PackageCard's issues - ie, a separate issue?

I'll try to review this later.

benogle:
I love this idea. I think we can fix the issues @mnquintana pointed out in other PRs, then merge this.

benogle:
Merged in #572

PR #541, opened 2015-06-11
:bug: Fix install-panel searching when reattached

### LINKED ISSUE #535: Install: searching after closing pane doesn't work on enter press
Steps to reproduce:

1. Open settings
2. Install
3. Search for anything and press enter to run the search
4. Items come up
5. Repeat 3-4 if you want to to prove that it'll work again and again
6. Close settings tab
7. Open settings (it should still be on install pane)
8. Search for anything and press enter to run the search
9. It doesn't work

As a side note, you can make it work by toggling to themes and then back to packages. So presumably the keybinding gets lost when the tab is first closed.

I'm using Atom 2.0.7.

Screencap (though not overly helpful as you'l have to trust that I'm pressing enter after typing my search terms!):

settings

azoerb:
Fixes #535

### DIFF
--- a/lib/install-panel.coffee
+++ b/lib/install-panel.coffee
@@ -62,6 +62,16 @@ class InstallPanel extends View
 
     @loadFeaturedPackages()
 
+  subscribeToEvents: ->
+    @subscribe @packageManager, 'package-install-failed', (pack, error) =>
+      @searchErrors.append(new ErrorView(@packageManager, error))
+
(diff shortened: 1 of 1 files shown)


mnquintana:
Would you mind adding specs for this to prevent regressions? Thanks!

azoerb:
Sure thing!

I'm trying to figure out how to test closing the settings screen without completely removing and re-adding it as in the other examples.  Is that the correct way to do it, or am I just missing how to do this?

Thanks

PR #552, opened 2015-06-25
Remove text related to deprecated APIs being removed in V1.0.

cgcardona:
Now that Atom is at V1.0 we no longer need to show a warning message regarding deprecated APIs being removed in the V1.0 release.

screen shot 2015-06-25 at 10 00 40 am


kevinsawicki:
Thanks!

PR #557, opened 2015-06-29
Listening for enter key on search input field

### LINKED ISSUE #553: search packages can't work. only after switching to themes and return to packages
Use case:
 Open Packages -> Settings view -> Install packages, then enter some text in search field, but nothing doing.
 for get result I must change Packages button to Themes and switch again for a get search result.
 No 'Enter' key action here too

dijs:
Fixes #553

### DIFF
--- a/lib/install-panel.coffee
+++ b/lib/install-panel.coffee
@@ -60,8 +60,6 @@ class InstallPanel extends View
     @searchType = 'packages'
     @handleSearchEvents()
 
-    @searchEditorView.on 'keyup', (e) => @performSearch if e.key is 13
-
     @loadFeaturedPackages()
 
   detached: ->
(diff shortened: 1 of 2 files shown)


dijs:
Is there any way to test this module locally?

thedaniel:
A spec for this would be nice.

If you run `apm link -d` from your checkout the linked version will override the bundled on in dev mode. You might need to run `apm rebuild` as well.

kevinsawicki:
Isn't enter already handled here: https://github.com/atom/settings-view/blob/e63a52c41342819b3b78c894a2e63f13391fbd16/lib/install-panel.coffee#L75-L76 ?

dijs:
@kevinsawicki The search method was not being called when I pressed enter in the input field until I added this line. Possibly a OS specific issue? I am on OSX.

PR #564, opened 2015-07-01
Fix several bugs and regressions

### LINKED ISSUE #535: Install: searching after closing pane doesn't work on enter press

### LINKED ISSUE #545: Can't uninstall deprecated package
I have the language-typescript package installed. Apparently it's deprecated and superseded by atom-typescript, but I'm not interested in keeping it around and would prefer to just uninstall it, but that option is missing for this package specifically.

screenshot from 2015-06-12 01 07 00

### LINKED ISSUE #556: Setting changes in config not updated in settings view
If I make changes in the config.cson file directly instead of the settings view, these settings will not be updated in the view. For example if I change the font size in the config.cson it will either remain default in the view or be the previous setting:

settings_bug

### LINKED ISSUE #559: Install Packages search not working
Saw a tweet that the search for new packages doesn't work.

tmlrhcw

From the gif it seems that the first search was successful, otherwise there would be a bigger gap after the input. Search term was probably `set`.

Can't reproduce it myself.

### LINKED ISSUE #562: Install toggle buttons should always perform a search
Carried over from https://github.com/atom/atom/issues/7590.

After entering a search term you could hit <kbd>enter</kbd> to perform a search, but some people might try to click the "packages/themes" toggle buttons. The problem is that it doesn't work for the already active button. Example below: Can't click on "Packages" and need to switch to "Themes" and back.

install

benogle:
Closes #559
Closes #557
Closes #562
Closes #556 with f47609116be0453a481e0cb9044e48c87d3a73d3
Closes #545
Closes #535
Closes #541

### DIFF
--- a/lib/general-panel.coffee
+++ b/lib/general-panel.coffee
@@ -11,5 +11,16 @@ class GeneralPanel extends View
   initialize: ->
     @loadingElement.remove()
 
-    @append(new SettingsPanel('core'))
-    @append(new SettingsPanel('editor'))
+    @subPanels = [
+      new SettingsPanel('core')
+      new SettingsPanel('editor')
+    ]
+
+    for subPanel in @subPanels
+      @append(subPanel)
+    return
+
+  dispose: ->
+    for subPanel in @subPanels
+      subPanel.dispose()
+    return
--- a/lib/install-panel.coffee
+++ b/lib/install-panel.coffee
@@ -68,23 +68,19 @@ class InstallPanel extends View
     @searchEditorView.focus()
 
   handleSearchEvents: ->
-    console.log 'setup events'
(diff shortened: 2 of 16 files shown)


izuzak:
:heart_eyes:  @benogle

Issue #550, opened 2015-06-23
Packages linked to ~/.atom/packages don't display Readmes


mnquintana:
Steps to reproduce:

1. Clone a package.
1. `cd` into that directory.
1. `apm link`
1. Reload Atom
1. Open up the package detail view for that package in Settings View
1. Scroll to the Readme section
1. There will be the following error: `ERROR: No README data found!`

mertzt89:
confirmed. I am experiencing the same issue.

t9md:
I quickly checked issue,  its seem [this line](https://github.com/atom/settings-view/blob/master/lib/package-detail-view.coffee#L116) won't never null, since `@pack.metadata.readme`  set to 'ERROR: No README data found!!'.
So I need to investigate where `metadata.readme` is set.

t9md:
From atom v0.199,  packages metadata become [normalized](https://github.com/atom/atom/commit/97ad2f30e028eb24320f86390aadb3a4b8bd09d2) with `normalize-package-data` and it set (ERROR: No README data found!)[https://github.com/npm/normalize-package-data/blob/2b006a6d53490fc053ba5ee727d36fbb4ff64868/lib/fixer.js#L227-L232] string.

By this change, pack.metadata.readme always set `string` data.
So [this line](https://github.com/atom/settings-view/blob/master/lib/package-detail-view.coffee#L116-L117) never be true, since readme is never be set `null`.


I can't judge where (atom side or settings.views specific logic) to be fixed.
Hope this investigation helps to get fixed.

nemoDreamer:
:+1:

rrd108:
I have the same issue with my own package.

nemoDreamer:
:lollipop:'s for @mnquintana !!!

PR #566, opened 2015-07-03
Display Readmes for symlinked packages

mnquintana:
Fixes https://github.com/atom/settings-view/issues/550

/cc @atom/feedback


thedaniel:
Looks good to me!

jlord:
:+1:

t9md:
Thanks

For future resilience to the change of normalize-package-data module, its better to fix atom/atom side like keeping original readme field before normalize and restore it after normalization.
This approach don't depend on special `'ERROR: No README data found!'` string.

But anyway I'm happy to fix this immediately rather than waiting for atom/atom side fix.

Issue #348, opened 2015-01-20
Some packages add a spurious ".bin" package to settings view


smlombardi:
Soon after the new settings view was included, I noticed this odd item:

![screen shot 2015-01-19 at 2 40 11 pm](https://cloud.githubusercontent.com/assets/835148/5810676/3dce3714-a018-11e4-980e-30126bab9d21.png)

Which seems to be related to autoprefixer package.  See also https://discuss.atom.io/t/whats-this-bin-package/14355/7 for the initial thread.

ggamel:
I'm experiencing this same issue. Very strange.

k-j-kleist:
I'm seeing this too (Atom 0.198.0, Windows 7).

I do have a ".bin" directory in "%USERPROFILE%\.atom\packages" containing two files:
```
"atbuild":

"$basedir/../atom-typescript/dist/main/bin/build.js"   "$@"
exit $?

"atbuild.cmd":

"$basedir/../atom-typescript/dist/main/bin/build.js"   "$@"
exit $?
```
I installed the package "atom-typescript" some days ago, but have installed several ones after it.

atrefz:
:+1:

rjmunro:
Can the title of this bug be changed as it's not related to Autoprefixer specifically. It happens with any plugin that includes executable scripts/binaries to run from the command line. It's standard behaviour that has come from npm. Either:

* The UI should ignore package folders with names that start with `.`
* apm should be configured to put the bin symlinks somewhere else (maybe `~/.atom/bin`)
* Packages should be banned from including these kinds of scripts and warnings / errors should be shown when they are installed.

obiwanakin:
I also have this issue

madcampos:
this is observable in 1.0.2 on Windows 10

cebor:
same issue here on osx with atom 1.0.3

mauron85:
+1

```bash
ls -l .atom/packages/.bin/
```

```
atbuild@ -> ../atom-typescript/dist/main/bin/atbuild
```

impinball:
Confirmed on Ubuntu 14.04 64-bit, Atom 1.0.11. In my case, it's also caused by atom-typescript. BTW packages starting with a leading dot aren't valid packages to begin with, so they should be ignored.

thedaniel:
> - apm should be configured to put the bin symlinks somewhere else (maybe ~/.atom/bin)

[...]

PR #569, opened 2015-07-05
Hide packages that start with a dot

helarqjsc:
Refs #348 
It would probably be a better solution to not put .bin folder in the package folder to begin with, but as a temporary solution it should be hidden from the list at least.

Also, I haven't added any new specs, because it already takes the filters into account. I've added a package with name ".bin" to the fixtures and the tests are being passed properly. The spec "InstalledPackagesPanel shows packages" is hard-coded to expect only a single community package to be displayed, so it already checks that the package with name ".bin" is filtered out.


thedaniel:
This is reasonable, but it should happen in https://github.com/atom/apm since that's where the package list is initially generated (not to mention that `.bin` shows up in `apm list` as well)

Issue #531, opened 2015-06-07
Use default cursor


jerone:
Showing the cursor as a "hand" when hovering over things in the UI is a very "web" oriented convention. Both the tabs and tree-view package use the default cursor. I would propose that the settings-view does the same, and any other package for that matter.

jerone:
In an attempt to clean up my created issues that have not been processed or updated over a year, I'm closing this issue. If this issue is still relevant, please reopen another issue.

PR #575, opened 2015-07-07
Avoid use of cursor: pointer

mnquintana:
Depends on https://github.com/atom/atom/pull/7698
Closes https://github.com/atom/settings-view/issues/531

This replaces usage of `cursor: pointer` with `cursor: default` whenever possible, except for external links. This also adds new styling for internal links (which will probably need some tweaking, especially for colors) – @simurai I experimented with a `background-color` highlight for internal links [here](https://github.com/atom/atom/pull/7698/files#diff-16c6799dcc54b3fb4e5ed198c28d06a1R17).

This PR also extracts a lot of component Less into separate files, as the main Less file was starting to get really messy.

(Be sure to only test with the branch from https://github.com/atom/atom/pull/7698, as it depends on styles added there.)

/cc @atom/feedback


damieng:
This has been open for quite some time - worth keeping around or is it dead?

mnquintana:
I think it's probably still worth doing at some point, but I don't wanna think about resolving the merge conflicts from this PR :see_no_evil: so I'm gonna close this out for now

simurai:
> @mnquintana I think it's probably still worth doing at some point

Yeah, I also think it's still a good idea. Maybe for now it can be styled directly one by one and not introduce new classes. Then PRs don't need to be synced up everywhere. I'll keep in mind.

PR #582, opened 2015-07-10
Remove redundant updates container on the package detail view

mnquintana:
Now that updates are displayed inline in package cards, the available updates container above the package card in the package detail view is a bit redundant - this removes it.

### Before

settings_-__users_machistequintana_github_atom-dark-ui_-_atom

### After

settings_-__users_machistequintana_github_settings-view_-_atom

/cc @benogle @thedaniel @simurai

### DIFF
--- a/lib/package-detail-view.coffee
+++ b/lib/package-detail-view.coffee
@@ -30,10 +30,6 @@ class PackageDetailView extends View
 
       @section class: 'section', =>
         @form class: 'section-container package-detail-view', =>
-          @div outlet: 'updateArea', class: 'alert alert-success package-update', =>
-            @span outlet: 'updateLabel', class: 'icon icon-squirrel update-message'
-            @span outlet: 'updateLink', class: 'alert-link update-link icon icon-cloud-download', 'Install'
-
           @div class: 'container package-container', =>
             @div class: 'row', =>
               @subview 'packageCard', new PackageCard(pack.metadata, packageManager, onSettingsView: true)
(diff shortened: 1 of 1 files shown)


benogle:
Nice, thanks. Yay for removal

lib/package-detail-view.coffee
```diff
@@ -201,30 +196,21 @@ class PackageDetailView extends View
     loadTime + activateTime
 
   installUpdate: ->
```
mnquintana: @benogle I think we might be able to :fire: this whole method since the package card is now responsible for installing updates (and handling errors?) - what do you think?

benogle: Sounds good. I assume nothing is calling this now. Remove any dead code you can.

mnquintana:
@benogle Alright, removed a bunch of dead code now. :ship:?

benogle:
Where is the update check happening now? Might need to keep in the outdated check for now, then update the card with the new version. 

What would be really nice (and something I have on my list) is for the outdated fetching to emit an event that the rest of the interface listens to.

benogle:
I might be trippin though, and it may be happening already

mnquintana:
> Where is the update check happening now? Might need to keep in the outdated check for now, then update the card with the new version.

I think it happens when `PackageCard` is initialized in [this call to `updateInterfaceState()`](https://github.com/atom/settings-view/blob/mq-rm-pack-update-box/lib/package-card.coffee#L92), so nothing should be dependent on `PackageDetailView::checkForUpdate()`

benogle:
ok sounds good

benogle:
ok sounds good

PR #589, opened 2015-07-15
Support descriptions for enum values in config.

oggy:
This goes with atom/atom#7916

### DIFF
--- a/spec/settings-panel-spec.coffee
+++ b/spec/settings-panel-spec.coffee
@@ -31,22 +31,36 @@ describe "SettingsPanel", ->
             type: 'string'
             default: 'zang'
             order: 100
+          enum:
+            title: 'An enum'
+            type: 'string'
+            default: 'one'
+            enum: [
+              {value: 'one', description: 'One'}
+              'Two'
+            ]
       atom.config.setSchema("foo", config)
       atom.config.setDefaults("foo", gong: 'gong')
-      expect(_.size(atom.config.get('foo'))).toBe 5
+      expect(_.size(atom.config.get('foo'))).toBe 6
       settingsPanel = new SettingsPanel("foo", {includeTitle: false})
 
     it "sorts settings by order and then alphabetically by the key", ->
       settings = atom.config.get('foo')
-      expect(_.size(settings)).toBe 5
+      expect(_.size(settings)).toBe 6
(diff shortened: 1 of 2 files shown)


spec/settings-panel-spec.coffee
```diff
@@ -31,22 +31,36 @@ describe "SettingsPanel", ->
             type: 'string'
             default: 'zang'
             order: 100
+          enum:
+            title: 'An enum'
+            type: 'string'
+            default: 'one'
+            enum: [
+              ['one', 'One']
```
benogle: What about an object for clarity? `{value: 'one', description: 'One'}`

thomasjo: :+1: Makes it even clearer and sets up for making it easier to possibly adding further metadata in the future.

thomasjo:
Specs failures seemed legit, but I kicked off a new build just to be on the safe side.

benogle:
Likely legit cause behavior is not in master.

oggy:
I've updated this and the other PR to use objects with `value` and `description` properties.

I think the tests are failing because the updated test here depends on the other PR being merged, and the atom on Travis would not include that. If I run locally with `ATOM_DEV_RESOURCE_PATH` pointing to the PR branch of Atom, it passes.

Is that ok? Do we need to have a test suite that works with older Atoms? I figured being a core package this was fine, but let me know if not.

thomasjo:
> I've updated this and the other PR to use objects with value and description properties.

Awesome! :clap: 

> Is that ok? Do we need to have a test suite that works with older Atoms? I figured being a core package this was fine, but let me know if not.

I think that's totally OK. Nobody should ever install this manually anyway (outside of hacking on it), so that isn't really something we need to handle.

This looks good to me! :ship:

Arcanemagus:
Is there a reason the associated Atom PR was merged back on 2016-01-26 but this was never merged?

lee-dohm:
> Is there a reason the associated Atom PR was merged back on 2016-01-26 but this was never merged?

Someone made a mistake? (These things happen, y'know :laughing:)

I'm going to take a look at this and see if it can easily be updated and merged. Not sure how much drift there has been since it was last looked at.

lee-dohm:
Closing this as it is superseded by #788.

lee-dohm:
Thanks very much @oggy for the work you did!

PR #593, opened 2015-07-18
Collapse packages/themes on click

### LINKED ISSUE #475: Collapse package section on title click
Request: Click on the community packages section (or any other section) in the installed packages panel and have that section collapse. 

Use Case: There are a large number of packages installed and the user is looking for a particular package that they don't remember the name to but know what section it's in. 

(Clicking what I've underlined here would collapse the section)

image

acontreras89:
Minimal implementation of the enhancement requested in #475.

- [x] Make packages and themes to collapse/expand on click
- [x] Make dynamically added entries to be shown or hidden depending on the visibility of the list (e.g. when installing a new package or theme)
- [x] Add specs

Closes #475

### DIFF
--- a/lib/installed-packages-panel.coffee
+++ b/lib/installed-packages-panel.coffee
@@ -208,4 +208,4 @@ class InstalledPackagesPanel extends View
 
   handleEvents: ->
     @on 'click', '.sub-section .icon-package', (e) ->
-      [].map.call(e.currentTarget.nextSibling.children, (child) -> child.hidden = not child.hidden)
+      e.currentTarget.parentNode.classList.toggle('collapsed')
--- a/lib/installed-packages-panel.coffee
+++ b/lib/installed-packages-panel.coffee
@@ -81,7 +81,6 @@ class InstalledPackagesPanel extends View
       , InstalledPackagesPanel.loadPackagesDelay
 
     @handleEvents()
-
     @loadPackages()
 
   focus: ->
--- a/lib/installed-packages-panel.coffee
+++ b/lib/installed-packages-panel.coffee
@@ -80,6 +80,8 @@ class InstalledPackagesPanel extends View
         @loadPackages()
       , InstalledPackagesPanel.loadPackagesDelay
 
+    @handleEvents()
+
     @loadPackages()
 
   focus: ->
@@ -203,3 +205,7 @@ class InstalledPackagesPanel extends View
   matchPackages: ->
     filterText = @filterEditor.getModel().getText()
     @filterPackageListByText(filterText)
+
+  handleEvents: ->
+    @on 'click', '.sub-section .icon-package', (e) =>
+      _.map(e.currentTarget.nextSibling.children, (child) -> child.hidden = !(child.hidden))
--- a/lib/installed-packages-panel.coffee
+++ b/lib/installed-packages-panel.coffee
@@ -208,4 +208,4 @@ class InstalledPackagesPanel extends View
 
   handleEvents: ->
     @on 'click', '.sub-section .icon-package', (e) ->
-      _.map(e.currentTarget.nextSibling.children, (child) -> child.hidden = not child.hidden)
+      [].map.call(e.currentTarget.nextSibling.children, (child) -> child.hidden = not child.hidden)
--- a/lib/installed-packages-panel.coffee
+++ b/lib/installed-packages-panel.coffee
@@ -207,5 +207,5 @@ class InstalledPackagesPanel extends View
     @filterPackageListByText(filterText)
 
   handleEvents: ->
-    @on 'click', '.sub-section .icon-package', (e) =>
-      _.map(e.currentTarget.nextSibling.children, (child) -> child.hidden = !(child.hidden))
+    @on 'click', '.sub-section .icon-package', (e) ->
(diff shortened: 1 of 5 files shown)


50Wliu:
This will need some specs.

lib/themes-panel.coffee
```diff
@@ -299,3 +301,7 @@ class ThemesPanel extends View
   matchPackages: ->
     filterText = @filterEditor.getModel().getText()
     @filterPackageListByText(filterText)
+
+  handleEvents: ->
+    @on 'click', '.sub-section .icon-paintcan', (e) ->
+      _.map(e.currentTarget.nextSibling.children, (child) -> child.hidden = not child.hidden)
```
mnquintana: Any reason you're using `underscore-plus`'s `map` function here instead of native `Array.map`?

acontreras89: No reason. I'll change that right away

acontreras89:
Specs added @50Wliu

thomasjo:
This looks really good! Maybe we can have one spec that only collapses one sub-section? That could help guard against any future regressions. Besides that one minor nitpick, this looks really to :boat:

lib/installed-packages-panel.coffee
```diff
@@ -80,6 +80,8 @@ class InstalledPackagesPanel extends View
         @loadPackages()
       , InstalledPackagesPanel.loadPackagesDelay
 
+    @handleEvents()
+
```
thomasjo: :fire: empty line

lib/themes-panel.coffee
```diff
@@ -88,6 +88,8 @@ class ThemesPanel extends View
       core: new ListView(@items.core, @corePackages, @createPackageCard)
       user: new ListView(@items.user, @communityPackages, @createPackageCard)
 
+    @handleEvents()
+
```
thomasjo: :fire: empty line

acontreras89:
Updated the specs and removed the empty lines as you suggested, @thomasjo

styles/settings-view.less
```diff
@@ -290,6 +290,10 @@
     margin: @section-padding 0;
   }
 
+  .sub-section.collapsed .package-container .row {
```
mnquintana: I think this selector could be simplified to just `.sub-section.collapsed .package-container` – I'd also like to avoid referencing Bootstrap classes (like `.row`) since this package will stop using the Bootstrap grid at some point.

acontreras89: Using `.sub-section.collapsed .package-container` makes sections with zero packages to change in size when their headers are clicked. Would that be ok?

mnquintana: Hmmm what about `.sub-section.collapsed .package-container .package-card`?

acontreras89: Sure, no problem! :dancer:

thomasjo:
Been trying this out this morning and I like it. I do think the affordance of the click to collapse/expand is missing. There's no hint that clicking will expand or collapse the section. We will have to fix that. The easiest route is to use the *fold* and *unfold* icons. I'll experiment with this shortly.

acontreras89:
Thanks @thomasjo 
I played a little bit with some icons, but couldn't find anything that felt right (not really good at styling, tbh).
That's why I described the PR as a **minimal** implementation of the enhancement :stuck_out_tongue: 

Let me know if you need anything!

thomasjo:
@acontreras89 Here's what I came up with: https://github.com/atom/settings-view/commit/4cc717d71c460cb054fe2a0d7b093b975e77a810

settings-view-tweaks

You should be able to cherry-pick that commit into your branch (if you like it). I can tweak it further first if you want. And if you don't like, be honest.

thomasjo:
By the way, is there any reason you're targeting `.package-card` elements for the collapsing? Why not just remove that from the selector and hide the entire `.package-container` instead? In other words change the selector to this
```less
.sub-section.collapsed .package-container {
  display: none;
}
```

There might be a very good reason that I'm completely ignorant to, just figured I'd ask the obvious question.

thomasjo:
I also think the waster whitespace between collapsed sections should be altered, but I can fix that later in a separate PR, just mentioning it for posterity.

mnquintana:
@thomasjo Looks good to me! :+1:

> By the way, is there any reason you're targeting .package-card elements for the collapsing?

@thomasjo https://github.com/atom/settings-view/pull/593#discussion_r36048751

thomasjo:
@mnquintana Thanks for the heads up. We can tweak this little things later. There are some improvements I'd like to make, but I don't want that to stop this PR from moving forward. This is already a huge improvement over the current situation.

acontreras89:
> There are some improvements I'd like to make, but I don't want that to stop this PR from moving forward. This is already a huge improvement over the current situation.

I definitely agree.
Give me 20 minutes and I'll push your commit :)

thomasjo:
Unless anyone has any further comments or suggestions, I think this is ready to :ship:.

/cc @atom/feedback

benogle:
:+1:

thomasjo:
Thanks @acontreras89! Awesome work :heart:

acontreras89:
No problem at all, @thomasjo 
I'll keep it coming as long as you guys make it this easy :ok_hand:

simurai:
Nice.. :clap:

PR #610, opened 2015-07-29
Add toggle for keymap activation in package settings

tmunro:
Adds UI support for toggling keymap activation on a package-level. See https://github.com/atom/atom/pull/8130 for more details.

screenshot 2015-07-29 08 58 40

### DIFF
--- a/lib/package-keymap-view.coffee
+++ b/lib/package-keymap-view.coffee
@@ -1,6 +1,7 @@
 path = require 'path'
 _ = require 'underscore-plus'
 {$, $$$, View} = require 'atom-space-pen-views'
+KeybindingsPanel = require './keybindings-panel'
 
 # Displays the keybindings for a package namespace
 module.exports =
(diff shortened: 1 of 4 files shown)


benogle:
Can you paste in a screenshot for this?

tmunro:
Absolutely, added screenshot.

maxbrunsfeld:
Ok, the dependent atom/atom PR is merged, so we can merge this once Atom 1.0.4 has shipped.

This looks good to me UI-wise, though I'd probably tweak the description line (`Whether to use the keybindings from this package`), or possibly just remove that. Maybe @benogle or @thedaniel would have some ideas for copy on this.

@tmunro Could you add a spec for this new button? You could just assert that the package's name is added/removed from the `core.disabledKeymaps` config.

Thanks!

tmunro:
Fixed the broken specs and added that new one. One thing I don't quite understand is why calling my new `hasKeymaps` method was breaking the specs. I inlined it for now, but it doesn't really make sense to me.

lib/package-keymap-view.coffee
```diff
@@ -8,21 +9,53 @@ class PackageKeymapView extends View
   @content: ->
     @section class: 'section', =>
       @div class: 'section-heading icon icon-keyboard', 'Keybindings'
+      @div class: 'checkbox', =>
+        @label for: 'toggleKeybindings', =>
+          @input id: 'toggleKeybindings', type: 'checkbox', outlet: 'keybindingToggle'
+          @div class: 'setting-title', 'Use Keybindings?'
```
mnquintana: What do you think about dropping the `?` here? I think we tend to avoid punctuation when rendering setting names.

maxbrunsfeld: :+1:

tmunro:
Removed the ? from the setting name and changed the description to help explain why you might want to use the setting.

maxbrunsfeld:
OK @tmunro, just a heads up that I've merged atom/atom#8173. Sorry for the delay on that one; I was out on vacation.

tmunro:
I think this is ready to merge, now that Atom 1.0.8 is out it shouldn't break anything.

thomasjo:
Not sure I love the the title of the setting (*Enable*), but someone can always change that later. I'm gonna merge and :ship: this.

Thanks so much for this @tmunro! :heart: :green_heart: :yellow_heart:

PR #614, opened 2015-07-31
Package show URI

thedaniel:
This allows a package detail view to be shown directly from a URI, whether it's installed or not. This is useful for e.g. installing packages from links on the webiste (atom/atom#5262) or linking directly to the the metrics package from atom/welcome to make it simple to disable (atom/atom#4966) among other things. 

Here's a dump of my notes for the flow, it should match the specs as well.

```
# PackageDetailView:

- Is pack present?
  - Yes
    - is it complete?
      - Yes
        - Does everything render normally? Cool.
      - No
        - [A] Is the network available?
          - Yes
            - Call the network.
              - Failure?
                - Show an error page
              - Success?
                - Render that thing (don't forget to rerender the subview)
  - No
    - See [A]
```

### DIFF
--- a/lib/package-card.coffee
+++ b/lib/package-card.coffee
@@ -174,9 +174,11 @@ class PackageCard extends View
       @avatar.attr 'src', "file://#{avatarPath}" if avatarPath
 
     @client.package @pack.name, (err, data) =>
-      data ?= {}
-      @packageData = data
-      @downloadCount.text data.downloads?.toLocaleString()
+      if err
+        # Just don't update the download count
+      else
+        data ?= {}
+        @downloadCount.text data.downloads?.toLocaleString()
 
   updateInterfaceState: ->
     @versionValue.text(@installablePack?.version ? @pack.version)
--- a/lib/package-card.coffee
+++ b/lib/package-card.coffee
@@ -174,9 +174,9 @@ class PackageCard extends View
       @avatar.attr 'src', "file://#{avatarPath}" if avatarPath
 
     @client.package @pack.name, (err, data) =>
-      if err
-        # Just don't update the download count
-      else
+      # We don't need to actually handle the error here, we can just skip
+      # showing the download count if there's a problem.
+      unless err
         data ?= {}
         @downloadCount.text data.downloads?.toLocaleString()
 
--- a/spec/package-detail-view-spec.coffee
+++ b/spec/package-detail-view-spec.coffee
@@ -1,3 +1,4 @@
+fs = require 'fs'
 path = require 'path'
 
 PackageDetailView = require '../lib/package-detail-view'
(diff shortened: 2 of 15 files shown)


thedaniel:
Anywho this PR represents the design decision that `PackageDetailView` is responsible for figuring out whether the whole package is there and hitting the network and displaying the right state. An alternative is handling it in `main.coffee` or `SettingsView` but this feels right to me.

thomasjo:
Can't wait! Thanks for working on this @thedaniel.

thedaniel:
@atom/feedback i think this is ready to review. I'm not fully confident that it handles all error states but I've tried to handle network errors in a simple, catch-all kind of way.

I'd like to leave this undocumented for now as the initial use is just intended to be https://github.com/atom/welcome/pull/38 - you can link that branch in dev mode to see it in action from the welcome package.

When we address https://github.com/atom/atom/issues/5262 we may discover some necessary changes, etc, and I would suggest documenting this at that time.

lib/package-card.coffee
```diff
@@ -174,9 +174,11 @@ class PackageCard extends View
       @avatar.attr 'src', "file://#{avatarPath}" if avatarPath
 
     @client.package @pack.name, (err, data) =>
-      data ?= {}
-      @packageData = data
-      @downloadCount.text data.downloads?.toLocaleString()
+      if err
```
benogle: prolly should use an `unless` here

spec/package-detail-view-spec.coffee
```diff
@@ -0,0 +1,71 @@
+fs = require 'fs'
```
benogle: yay for adding specs to these views

spec/package-detail-view-spec.coffee
```diff
@@ -0,0 +1,71 @@
...
+    packageManager.client = createClientSpy()
+    packageManager.client.package.andCallFake (name, cb) ->
+      packageData = require(path.join(__dirname, 'fixtures', 'package-with-readme', 'package.json'))
+      packageData.readme = fs.readFileSync(path.join(__dirname, 'fixtures', 'package-with-readme', 'README.md'), 'utf8')
+      cb(null, packageData)
+    view = new PackageDetailView({name: 'package-with-readme'}, packageManager)
+    view.beforeShow(opts)
+
+
+  it "Renders a package when provided in `initialize`", ->
```
benogle: I think we lowercase the beginning of these.

spec/package-detail-view-spec.coffee
```diff
@@ -0,0 +1,71 @@
...
+    opts ?= {}
+    packageManager.client = createClientSpy()
+    packageManager.client.package.andCallFake (name, cb) ->
+      packageData = require(path.join(__dirname, 'fixtures', 'package-with-readme', 'package.json'))
+      packageData.readme = fs.readFileSync(path.join(__dirname, 'fixtures', 'package-with-readme', 'README.md'), 'utf8')
+      cb(null, packageData)
+    view = new PackageDetailView({name: 'package-with-readme'}, packageManager)
+    view.beforeShow(opts)
+
+
```
benogle: :art: maybe one newline here?

benogle:
Will be cool to have this :+1:

thedaniel:
I'm going to merge this after our next release goes out, likely tuesday, and then get that metrics update in right after.

thomasjo:
Been testing this for a couple of days now, and not noticed any regressions or bugs :sparkles: 

Only thing missing, but that could perhaps be addressed in a separate PR, is serializing the current package panel details. If I e.g. reload Atom while viewing a package's config, I'll land back in the core settings panel after the reload. Whereas I expect to still be viewing the package config panel. At the very least I expect to remain in the *Packages* section.

thedaniel:
@thomasjo given that settings-view doesn't serialize that on master either (in other words, it's not a regression here), a separate issue / PR is fine.

PR #616, opened 2015-07-31
Fix bugs present when running Atom on Electron

thomasjo:
Basically something must have changed in Chromium 43+ with regards to how `HTMLSelectElement` behaves along with jQuery's `.prop()` function; using `.prop('selected', true)` no longer works as expected, so the workaround is to fall back to good ol' `.attr('selected', true)`.

Similarly e.g. `.val('atom-light-ui')` no longer works as a means of selecting an option, and the solution is use an equivalent workaround as with `.prop()`.

This might be a bug in jQuery and/or the Chromium build (43.0.2357.65) currently used by Electron, but at least this fixes the problem for the time being.

/cc @atom/feedback

### DIFF
--- a/lib/themes-panel.coffee
+++ b/lib/themes-panel.coffee
@@ -218,11 +218,11 @@ class ThemesPanel extends View
       switch metadata.theme
         when 'ui'
           themeItem = @createThemeMenuItem(name)
-          themeItem.prop('selected', true) if name is @activeUiTheme
+          themeItem.attr('selected', true) if name is @activeUiTheme
           @uiMenu.append(themeItem)
         when 'syntax'
           themeItem = @createThemeMenuItem(name)
-          themeItem.prop('selected', true) if name is @activeSyntaxTheme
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
:clap: Thanks for continuing to power through this.

mnquintana:
> This might be a bug in jQuery and/or the Chromium build (43.0.2357.65) currently used by Electron

It could be jQuery – we're using [jQuery 2.0.3](https://github.com/atom-archive/space-pen/blob/v5.0.1/vendor/jquery.js) via atom-space-pen-views – looking into it now

thomasjo:
@mnquintana Thanks buddy!

jlord:
😀😀😀

benogle:
> Thanks for continuing to power through this.

For real :heart: :green_heart: :blue_heart: :purple_heart: :yellow_heart:

mnquintana:
Looking through jQuery's release notes I didn't see anything that would affect this, so looks good to me :+1:

PR #622, opened 2015-08-07
Package README styling

### LINKED ISSUE #620: Markdown tables render incorrectly in package Readmes
Looks like Markdown tables currently don't render as expected in package Readmes:

settings_-__users_machistequintana_code_campus-map_-_atom

/cc @simurai

simurai:
This PR improves the styling of package READMEs

screen shot 2015-08-07 at 12 56 53 pm

- [x] `<table>` styling
- [x] `<kbd>` styling
- [x] more fixes like `a`, `hr`, `blockquote`, `pre`

Fixes #620

### DIFF
--- a/styles/package-readme.less
+++ b/styles/package-readme.less
@@ -15,11 +15,26 @@
       color: @text-color-highlight;
     }
 
+    a {
+      color: @text-color-info;
+    }
+
+    hr {
+      border-color: @text-color-subtle;
+    }
+
+    blockquote {
+      border-left-color: @text-color-subtle;
+    }
+
+    pre {
+      margin-bottom: 1em;
+    }
+
     img {
       max-width: 100%;
     }
 
-    // Table
     table {
       display: block;
       margin-bottom: 1em;
@@ -46,7 +61,6 @@
       }
     }
 
-    // Keybindings
     kbd {
       display: inline-block;
       padding: .3em .4em;
--- a/styles/package-readme.less
+++ b/styles/package-readme.less
@@ -19,7 +19,7 @@
       max-width: 100%;
     }
 
-    // Tables
+    // Table
     table {
       display: block;
       margin-bottom: 1em;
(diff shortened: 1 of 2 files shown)


simurai:
Empty cells in `<table>` don't get rendered. Maybe that shouldn't be a blocker for this PR?

jlord:
Interesting that it doesn't render empty cells. Not so _terrible_ if empty cells are on the end, but if they're in the middle then it could throw off the table. I found the issue for [this](https://github.com/chjj/marked/issues/476) in `marked`.

@gjtorikian we use your `roaster` for our markdown. Have you run into the problem with empty table cells being skipped by `marked` before? I think we could patch it by handling empty cells in the `table cell` [block level render methods](https://www.npmjs.com/package/marked#block-level-renderer-methods) in `marked`. What do you think? If it's not a terrible idea I can get the PR started.

gjtorikian:
I think that makes sense.

If the PR doesn't get merged we might also be able to getting away with some kind of hack using `&nbsp;` to insert a space.

jlord:
@simurai Turns out that the `settings-view` readme doesn't have _empty_ cells, but the raw markdown just didn't provide the any cells at all beyond the first two columns. GitHub.com fills these in apparently.

So at first, once I discovered that, I was like, _cool, there's actually no problem_. But then I realized, once I gave it the empty cells, it didn't render the last cell unless it had content. 

I used the `marked` render wrapper to insert content into blank cells, even tried the `&nbsp;` and that didn't help either, the last empty cell never gets included. Hmmm... 

screen shot 2015-08-11 at 12 54 51 pm

izuzak:
@jlord I think that's this issue: https://github.com/atom/markdown-preview/issues/206

jlord:
@izuzak ah, ok, thanks :sparkles: 

Then I think we should :ship: this PR @simurai for the table styles :+1:

PR #635, opened 2015-08-17
Allow Markdown in setting descriptions

### LINKED ISSUE #618: Render setting description links as links
link

We could probably render them as markdown.

lee-dohm:
Allows Markdown in setting descriptions, for example from my [tabs-to-spaces package](https://atom.io/packages/tabs-to-spaces) ...

**Before:**

screen shot 2015-08-17 at 7 46 26 pm

**After:**

screen shot 2015-08-17 at 7 47 13 pm

Fixes #618

### DIFF
--- a/lib/rich-description.coffee
+++ b/lib/rich-description.coffee
@@ -1,18 +1,12 @@
 marked = require 'marked'
 
 renderer = new marked.Renderer()
-renderer.code = ->
-  ''
-renderer.blockquote = ->
-  ''
-renderer.heading = ->
-  ''
-renderer.html = ->
-  ''
-renderer.image = ->
-  ''
-renderer.list = ->
-  ''
+renderer.code = -> ''
+renderer.blockquote = -> ''
+renderer.heading = -> ''
+renderer.html = -> ''
+renderer.image = -> ''
+renderer.list = -> ''
 
 markdown = (text) ->
   marked(text, renderer: renderer).replace(/<p>(.*)<\/p>/, "$1").trim()
--- a/lib/rich-description.coffee
+++ b/lib/rich-description.coffee
@@ -1,0 +1,23 @@
+marked = require 'marked'
+
+renderer = new marked.Renderer()
+renderer.code = ->
+  ''
+renderer.blockquote = ->
+  ''
+renderer.heading = ->
+  ''
+renderer.html = ->
+  ''
+renderer.image = ->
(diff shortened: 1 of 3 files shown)


lee-dohm:
There aren't any other specs for the Space Pen helpers in `settings-panel.coffee` and I wasn't able to easily pull the requisite methods up into the SettingsPanel class, so I left out the specs. Let me know if I need to get some specs in place to get this merged.

In the meantime, I'll continue trying.

thomasjo:
We definitely need to whitelist markup features; supporting headers, images, lists, HTML, and so forth, is a bad idea. It *will* be misused/abused. Off the top of my head, the only markup we should support is
 - `*italic*`
 - `**bold**`
 - `[link](http://atom.io)`
 - `` `code` ``

lee-dohm:
It looks like overriding the renderer might be a good method of whitelisting things:

https://www.npmjs.com/package/marked#overriding-renderer-methods

We can just have the elements that we don't want to support emit nothing. I think supporting all the inline methods except images and none of the block methods to begin with would be fine.

I'll extract the Markdown description stuff into a separate file to make it easy to spec. Then we can just call it from `settings-panel`.

thomasjo:
:boom: Nice!

lib/rich-description.coffee
```diff
@@ -0,0 +1,23 @@
...
+renderer.blockquote = ->
+  ''
+renderer.heading = ->
+  ''
+renderer.html = ->
+  ''
+renderer.image = ->
+  ''
+renderer.list = ->
+  ''
```
thomasjo: From a purely aesthetic viewpoint, it might look a bit more tidy if all of these were collapsed to single lines;
~~~ coffee
renderer.code = -> ''
renderer.blockquote = -> ''
renderer.heading = -> ''
renderer.html = -> ''
renderer.image = -> ''
renderer.list = -> ''
~~~

thomasjo:
Besides my :art: comment, I think this looks good :rocket:

jerone:
:+1: Great pull.

I'm wondering if blacklisted markdown should be stripped or kept as normal text.

lee-dohm:
@jerone I figured it was an easy first step to just strip it. If we find a use case where keeping it as normal text is really necessary (beyond "I didn't know that wasn't allowed") then we can add the complexity later.

Issue #632, opened 2015-08-14
Do some caching in loadOutdated()


benogle:
Currently `PackageManager::loadOutdated()` is stateless, and it sucks. Anytime any part of the settings view needs to know what outdated packages there are, it needs to hit the network and ask. This leads to https://github.com/atom/settings-view/issues/409. And this is being exacerbated by package cards in any view that can have the update information. No bueno. 

We should cache the outdated packages in the settings view's `PackageManager`. `loadOutdated` could do the network bit, load the cache, and emit events. This method could be called when it is currently for the status bar icon. Every consumer of the outdated data can call some other method like `getOutdated` which could return a promise, just in case the cache isnt loaded up yet.

This will fix https://github.com/atom/settings-view/issues/409, and could be the way to https://github.com/atom/settings-view/issues/615

cc @thedaniel

50Wliu:
As far as I can tell in my latest tests (with #887) there is caching going on in loadOutdated, and the updates panel loads _fast_.  As in, it takes longer to load the font icons then it does to load the package cards.

Closing.

PR #637, opened 2015-08-19
Cache results of ::loadOutdated

thedaniel:
~~~In progress. One of the specs fails when unfocused, and the cache isn't actually cleared when updating packages thru the API - maybe we're holding on to a bunch of instances of PackageManager or something. We can move the cache somewhere global if that's the case.~~~

Should be fine, I made a point to keep it pretty simple and testable.

cf. #632

### DIFF
--- a/lib/package-manager.coffee
+++ b/lib/package-manager.coffee
@@ -21,6 +21,8 @@ class PackageManager
         expiry: 0
 
     @emitter = new Emitter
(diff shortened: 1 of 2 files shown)


thedaniel:
Whoops, forgot to cc @atom/feedback. This is probably fine now.

thedaniel:
This is a pretty minor speedup but it's something. I guarantee there's something else in the check outdated -> display cards process begging for an optimization.

PR #643, opened 2015-08-24
remove margins from alert info

mertkahyaoglu:
<b>Before:</b>

before

<b>After:</b>

after


simurai:
This seems like good change. Thanks @mertkahyaoglu

mertkahyaoglu:
:+1:

Issue #652, opened 2015-09-13
Uncaught TypeError: path.extname is not a function


lee-dohm:
[Enter steps to reproduce below:]

1. Start Atom
1. Open the Settings View with <kbd>Cmd+,</kbd>
1. Search for a package that has key bindings
1. Scroll down to the key bindings section
1. Click the <span class="octicon octicon-clippy"></span> icon

**Atom Version**: 1.0.12-ad95281
**System**: Mac OS X 10.10.5
**Thrown From**: [settings-view](https://github.com/atom/settings-view) package, v0.216.0

### Stack Trace

Uncaught TypeError: path.extname is not a function

```
At /Applications/Atom.app/Contents/Resources/app.asar/node_modules/settings-view/lib/package-keymap-view.js:147

TypeError: path.extname is not a function
  at PackageKeymapView.module.exports.PackageKeymapView.writeKeyBindingToClipboard (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/settings-view/lib/package-keymap-view.js:147:30)
  at HTMLSpanElement.<anonymous> (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/settings-view/lib/package-keymap-view.js:139:24)
  at space-pen-section.jQuery.event.dispatch (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/jquery/dist/jquery.js:4435:9)
  at space-pen-section.elemData.handle (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/jquery/dist/jquery.js:4121:28)

```

### Commands

```
     -1:26.2.0 core:save (atom-text-editor.editor.is-focused)
     -1:25.8.0 core:backspace (atom-text-editor.editor.is-focused)
  2x -1:24.9.0 editor:newline (atom-text-editor.editor.is-focused)
     -1:22.6.0 core:backspace (atom-text-editor.editor.is-focused.autocomplete-active)
  2x -1:16.1.0 editor:newline (atom-text-editor.editor.is-focused)
  5x -1:12.8.0 core:backspace (atom-text-editor.editor.is-focused)
     -1:12.8.0 core:paste (atom-text-editor.editor.is-focused)
     -1:09.9.0 core:save (atom-text-editor.editor.is-focused)
 10x -1:08.3.0 editor:move-to-beginning-of-word (atom-text-editor.editor.is-focused)
  3x -1:05.6.0 editor:select-to-beginning-of-word (atom-text-editor.editor.is-focused)
[...]

Issue #648, opened 2015-09-04
Fail to copy the keybinding setting into clipboard 


leohacker:
Uncaught TypeError: undefined is not a function

[Enter steps to reproduce below:]

1. ...
2. ...

**Atom Version**: 1.0.11
**System**: Mac OS X 10.10.3
**Thrown From**: [settings-view](https://github.com/atom/settings-view) package, v0.216.0

### Stack Trace

Uncaught TypeError: undefined is not a function

```
At /Applications/Atom.app/Contents/Resources/app.asar/node_modules/settings-view/lib/package-keymap-view.js:147

TypeError: undefined is not a function
  at PackageKeymapView.module.exports.PackageKeymapView.writeKeyBindingToClipboard (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/settings-view/lib/package-keymap-view.js:147:30)
  at HTMLSpanElement.<anonymous> (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/settings-view/lib/package-keymap-view.js:139:24)
  at space-pen-section.jQuery.event.dispatch (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/jquery/dist/jquery.js:4435:9)
  at space-pen-section.elemData.handle (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/jquery/dist/jquery.js:4121:28)

```

### Commands

```
     -2:41.5.0 editor:cut-to-end-of-line (atom-text-editor.editor.is-focused)
     -2:40.3.0 editor:move-to-end-of-line (atom-text-editor.editor.is-focused.autocomplete-active)
     -2:39.3.0 editor:move-to-first-character-of-line (atom-text-editor.editor.is-focused.autocomplete-active)
     -2:38.9.0 core:move-up (atom-text-editor.editor.is-focused)
     -2:38.5.0 core:move-down (atom-text-editor.editor.is-focused)
     -2:38.2.0 editor:cut-to-end-of-line (atom-text-editor.editor.is-focused)
  5x -2:37.9.0 core:move-up (atom-text-editor.editor.is-focused)
     -2:36.8.0 editor:move-to-end-of-line (atom-text-editor.editor.is-focused)
     -2:33.7.0 fuzzy-finder:toggle-buffer-finder (atom-text-editor.editor.is-focused)
     -2:33 editor:consolidate-selections (atom-text-editor.editor.mini.is-focused)
[...]

PR #657, opened 2015-09-20
Don't overwrite path module variable

izuzak:
This fixes https://github.com/atom/settings-view/issues/648 and fixes https://github.com/atom/settings-view/issues/652. 

The reason why that `path.extname(...)` call was failing is that the `path` variable with the `require`d `path` Node module was being overwritten here. This fixes the problem by using a different name for the variable in the function.

Friendly cc to @tmunro @maxbrunsfeld @benogle since this code was added in https://github.com/atom/settings-view/pull/610 as far as I can tell. :v:

### DIFF
--- a/lib/package-keymap-view.coffee
+++ b/lib/package-keymap-view.coffee
@@ -42,7 +42,7 @@ class PackageKeymapView extends View
     @updateKeyBindingView()
 
     hasKeymaps = false
-    for [path, map] in atom.packages.getLoadedPackage(@namespace).keymaps
+    for [packageKeymapsPath, map] in atom.packages.getLoadedPackage(@namespace).keymaps
       if map.length > 0
         hasKeymaps = true
         break


tmunro:
Thanks! I noticed this error and had a sneaking suspicion it was my fault. Sorry about that!

It looks like the spec for copying keybindings only happens on the full keybindings panel, so that explains why I missed it at the time. Is it worth adding a duplicate spec just to check the copy button on the installed package view too?

izuzak:
> Is it worth adding a duplicate spec just to check the copy button on the installed package view too?

I think it is -- thanks for noticing that. Did you want to make a pull request to add those? If not, I'll make a note to add those (totally didn't think of that when fixing this :grin:).

tmunro:
I'll add the spec to prevent others from making my mistake :)

PR #660, opened 2015-09-21
Set User-Agent header when making network requests to GitHub/Atom.io

### LINKED ISSUE #640: No user-agent when requesting icons
When doing some log analysis for an internal GitHub project, I noticed a very large number of requests to `https://github.com/atom.png`. After clearing my settings-view cache in Atom and reloading the settings page, I immediately saw requests to this url from my IP (no one else was using Atom at the time), so I know it's the settings view that's loading these avatars. It does cache them after first load, but there's a lot of traffic for these still.

What made this particularly difficult to track down was that these avatar requests are coming in without a user agent header. Could you add something to the settings view to help identify these requests? For comparison, the GitHub desktop app uses `GitHubMac/31` or similar, so something like `Atom/1.0.7` would be appropriate.

Secondarily, I wouldn't mind if you more aggressively cached these - when loading the plugins from the atom org, it means repeated requests for the same avatar when only one request is needed.

cc @thedaniel

nathansobo:
Fixes #640

/cc @kevinsawicki

### DIFF
--- a/lib/atom-io-client.coffee
+++ b/lib/atom-io-client.coffee
@@ -3,6 +3,7 @@ path = require 'path'
 
 glob = null    # defer until used
 request = null # defer until used
+DefaultRequestHeaders = {'User-Agent': navigator.userAgent}
 
 module.exports =
 class AtomIoClient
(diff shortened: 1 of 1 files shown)


kevinsawicki:
:+1: Thanks for fixing this up /cc @zerowidth

Issue #655, opened 2015-09-19
Uncaught TypeError: Cannot convert undefined or null to object


gocarlos:
[Enter steps to reproduce below:]

1. ...
2. ...

**Atom Version**: 1.0.15
**System**: Microsoft Windows 10 Pro
**Thrown From**: [settings-view](https://github.com/atom/settings-view) package, v0.216.0

### Stack Trace

Uncaught TypeError: Cannot convert undefined or null to object

```
At C:\Users\gocarlos\AppData\Local\atom\app-1.0.15\resources\app.asar\node_modules\settings-view\lib\settings-view.js:273

TypeError: Cannot convert undefined or null to object
  at SettingsView.module.exports.SettingsView.getOrCreatePanel (C:\Users\gocarlos\AppData\Local\atom\app-1.0.15\resources\app.asar\node_modules\settings-view\lib\settings-view.js:273:44)
  at SettingsView.module.exports.SettingsView.showPanel (C:\Users\gocarlos\AppData\Local\atom\app-1.0.15\resources\app.asar\node_modules\settings-view\lib\settings-view.js:313:24)
  at openPanel (C:\Users\gocarlos\AppData\Local\atom\app-1.0.15\resources\app.asar\node_modules\settings-view\lib\main.js:41:25)
  at C:\Users\gocarlos\AppData\Local\atom\app-1.0.15\resources\app.asar\node_modules\settings-view\lib\main.js:69:13
  at Workspace.module.exports.Workspace.openURIInPane (C:\Users\gocarlos\AppData\Local\atom\app-1.0.15\resources\app.asar\src\workspace.js:446:22)
  at Workspace.module.exports.Workspace.open (C:\Users\gocarlos\AppData\Local\atom\app-1.0.15\resources\app.asar\src\workspace.js:379:19)
  at HTMLAnchorElement.<anonymous> (C:\Users\gocarlos\AppData\Local\atom\app-1.0.15\resources\app.asar\node_modules\about\lib\about-view.js:131:31)
  at space-pen-div.jQuery.event.dispatch (C:\Users\gocarlos\AppData\Local\atom\app-1.0.15\resources\app.asar\node_modules\jquery\dist\jquery.js:4435:9)
  at space-pen-div.elemData.handle (C:\Users\gocarlos\AppData\Local\atom\app-1.0.15\resources\app.asar\node_modules\jquery\dist\jquery.js:4121:28)

```

### Commands

```
```

### Config

```json
{
  "core": {}
}
```

[...]

PR #664, opened 2015-09-23
Prevent accessing panelCreateCallbacks before it is defined

izuzak:
This fixes https://github.com/atom/settings-view/issues/655.

e38bfa5 adds a failing test to demonstrate the problem from the issue. At that commit, running the specs outputs this error:

```
SettingsView
  when the package is activated
    when atom.workspace.open() is used with a config URI
      it opens the package settings view with atom://config/packages/<package-name>
        TypeError: Cannot convert undefined or null to object
          at SettingsView.module.exports.SettingsView.getOrCreatePanel (/Users/izuzak/github/settings-view/lib/settings-view.coffee:135:38)
```

11c5417 fixes the problem by preventing access to items of `panelCreateCallbacks` when it's not defined, similar to how it's done in other places in that function. Running the specs at this commit outputs 0 failures.

cc @kevinsawicki @thedaniel for :eyes:

### DIFF
--- a/lib/settings-view.coffee
+++ b/lib/settings-view.coffee
@@ -132,7 +132,7 @@ class SettingsView extends ScrollView
         panel = callback()
         @panelsByName ?= {}
         @panelsByName[name] = panel
-        delete @panelCreateCallbacks[name]
+        delete @panelCreateCallbacks?[name]
 
     panel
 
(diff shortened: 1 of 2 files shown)


kevinsawicki:
Looks great :ship:

Issue #634, opened 2015-08-15
Uncaught TypeError: Cannot read property 'on' of undefined


rpietro:
[Enter steps to reproduce below:]

1. ...
2. ...

**Atom Version**: 1.0.7
**System**: Mac OS X 10.10.4
**Thrown From**: [settings-view](https://github.com/atom/settings-view) package, v0.213.0

### Stack Trace

Uncaught TypeError: Cannot read property 'on' of undefined

```
At /Applications/Atom.app/Contents/Resources/app.asar/node_modules/settings-view/lib/package-card.js:303

TypeError: Cannot read property 'on' of undefined
  at PackageCard.module.exports.PackageCard.handleButtonEvents (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/settings-view/lib/package-card.js:303:23)
  at PackageCard.module.exports.PackageCard.initialize (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/settings-view/lib/package-card.js:191:12)
  at PackageCard.View (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/atom-space-pen-views/node_modules/space-pen/lib/space-pen.js:184:25)
  at new PackageCard (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/settings-view/lib/package-card.js:22:48)
  at PackageDetailView.module.exports.PackageDetailView.completeInitialzation (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/settings-view/lib/package-detail-view.js:148:28)
  at /Applications/Atom.app/Contents/Resources/app.asar/node_modules/settings-view/lib/package-detail-view.js:188:26
  at /Applications/Atom.app/Contents/Resources/app.asar/node_modules/settings-view/lib/atom-io-client.js:46:20
  at AtomIoClient.module.exports.AtomIoClient.fetchFromCache (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/settings-view/lib/atom-io-client.js:144:16)
  at AtomIoClient.module.exports.AtomIoClient.package (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/settings-view/lib/atom-io-client.js:43:19)
  at PackageDetailView.module.exports.PackageDetailView.fetchPackage (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/settings-view/lib/package-detail-view.js:180:56)
[...]

PR #666, opened 2015-09-25
Show error in PackageDetailView when network is unavailable and package is not in cache

izuzak:
This fixes https://github.com/atom/settings-view/issues/634. It seems that `AtomIoClient.fetchFromCache` calls the callback with [a special combination of values](https://github.com/atom/settings-view/blob/86699fbf3e237b6d659ed28990964c943d05f49a/lib/atom-io-client.coffee#L110-L113) in case the network is unavailable and the package was not found in cache. This special value was handled elsewhere, but needed to be handled in PackageDetailView as well.

 068ce24 adds a failing spec, which triggers this error:

```
PackageDetailView
  it shows an error when package metadata cannot be loaded from the cache and the network is unavailable
    TypeError: Cannot read property 'on' of undefined
      at PackageCard.module.exports.PackageCard.handleButtonEvents (/Users/izuzak/github/settings-view/lib/package-card.coffee:157:17)
      at PackageCard.module.exports.PackageCard.initialize (/Users/izuzak/github/settings-view/lib/package-card.coffee:69:6)
      at PackageCard.View (/Users/izuzak/github/settings-view/node_modules/atom-space-pen-views/node_modules/space-pen/lib/space-pen.js:184:25)
      at new PackageCard (/Users/izuzak/github/settings-view/lib/package-card.coffee:11:3)
      at PackageDetailView.module.exports.PackageDetailView.completeInitialzation (/Users/izuzak/github/settings-view/lib/package-detail-view.coffee:63:26)
      at /Users/izuzak/github/settings-view/lib/package-detail-view.coffee:98:10
      at /Users/izuzak/github/settings-view/lib/atom-io-client.coffee:32:9
      at AtomIoClient.<anonymous> (/Users/izuzak/github/settings-view/spec/package-detail-view-spec.coffee:69:7)
      at AtomIoClient.module.exports.AtomIoClient.package (/Users/izuzak/github/settings-view/lib/atom-io-client.coffee:30:6)
      at PackageDetailView.module.exports.PackageDetailView.fetchPackage (/Users/izuzak/github/settings-view/lib/package-detail-view.coffee:89:5)
      at PackageDetailView.module.exports.PackageDetailView.loadPackage (/Users/izuzak/github/settings-view/lib/package-detail-view.coffee:83:10)
      at PackageDetailView.module.exports.PackageDetailView.initialize (/Users/izuzak/github/settings-view/lib/package-detail-view.coffee:59:6)
      at PackageDetailView.View (/Users/izuzak/github/settings-view/node_modules/atom-space-pen-views/node_modules/space-pen/lib/space-pen.js:184:25)
      at new PackageDetailView (/Users/izuzak/github/settings-view/lib/package-detail-view.coffee:22:3)
      at [object Object].<anonymous> (/Users/izuzak/github/settings-view/spec/package-detail-view-spec.coffee:71:16)
```

d8cc61f fixes the problem by checking the data returned by `AtomIoClient` and verifying that it has at least the package's `name` property before showing the package card.

cc @kevinsawicki and @thedaniel for :eyes:

### DIFF
--- a/lib/package-detail-view.coffee
+++ b/lib/package-detail-view.coffee
@@ -87,7 +87,7 @@ class PackageDetailView extends View
   fetchPackage: ->
     @showLoadingMessage()
     @packageManager.getClient().package @pack.name, (err, packageData) =>
-      if err || !(packageData?.name?)
+      if err or not(packageData?.name?)
         @hideLoadingMessage()
         @showErrorMessage()
       else
--- a/lib/package-detail-view.coffee
+++ b/lib/package-detail-view.coffee
@@ -87,7 +87,7 @@ class PackageDetailView extends View
   fetchPackage: ->
     @showLoadingMessage()
     @packageManager.getClient().package @pack.name, (err, packageData) =>
-      if err
+      if err || !(packageData?.name?)
         @hideLoadingMessage()
         @showErrorMessage()
       else
--- a/spec/package-detail-view-spec.coffee
+++ b/spec/package-detail-view-spec.coffee
@@ -3,6 +3,7 @@ path = require 'path'
 
 PackageDetailView = require '../lib/package-detail-view'
 PackageManager = require '../lib/package-manager'
+AtomIoClient = require '../lib/atom-io-client'
 
 describe "PackageDetailView", ->
   packageManager = null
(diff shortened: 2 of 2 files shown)


thedaniel:
> AtomIoClient.fetchFromCache calls the callback with a special combination of values in case the network is unavailable and the package was not found in cache.

CLASSIC @thedaniel! Anyway, this seems fine for a quick fix, but if we're going to show an error, maybe we should just pass in an error instead of an imaginary value, and things will work in a more predictable way. Of course, that requires updating the other callers that special case `{}`

izuzak:
> Anyway, this seems fine for a quick fix, but if we're going to show an error, maybe we should just pass in an error instead of an imaginary value, and things will work in a more predictable way. Of course, that requires updating the other callers that special case `{}`

Yep, definitely agree with you. I'm intentionally going for a quick fix until I get a better understanding of this package's crazy complexity (I'm fixing small issues here and there for now). I'll open a new issue so that we don't forget to refactor this. :+1:

Issue #670, opened 2015-09-28
There shouldn't be a text input field for `core.customFileTypes`


maxbrunsfeld:
`customFileTypes` is a relatively new config field under the `core` schema. It is an object whose keys are scope names, and whose values are arrays of file extensions, like this:

```coffee
'*':
  core:
    customFileTypes:
      "source.cpp": [
        "h"
      ]
```

Settings view is rendering each key in the object as a text input, with the content `[object Object]`. This causes invalid data to be written to the config.

![screen shot 2015-09-28 at 11 59 58 am](https://cloud.githubusercontent.com/assets/326587/10145586/c8d71a8e-65d8-11e5-838a-1180bdc4f18d.png)

izuzak:
I've never looked at this part of code before, but here are some notes after looking at this for a bit. I suspect this will be related to the fact that the key of that property has a dot in it: `"source.cpp"`. My hunch is based on [this](https://github.com/atom/settings-view/blob/5a115f9e051da32e7af676daf6076d890d8d5e75/lib/settings-panel.coffee#L191) -> [this] (https://github.com/atom/atom/blob/8fb4e887df800438b2deeb23a2c23004b9ff358a/src/config.coffee#L639) -> [this](https://github.com/atom/atom/blob/8fb4e887df800438b2deeb23a2c23004b9ff358a/src/config.coffee#L1145-L1154). Basically, the dot is used as a namespace separator -- it has special meaning, so using it in the name of a property causes problems. If you use `"source-cpp"`, for example -- then everything works fine. Looking at that [last piece of code](https://github.com/atom/atom/blob/8fb4e887df800438b2deeb23a2c23004b9ff358a/src/config.coffee#L1150), it _seems_ as if there's a way to escape the dot?

PR #672, opened 2015-09-29
Don't render input fields for core.customFileTypes

maxbrunsfeld:
Fixes https://github.com/atom/settings-view/issues/670

### DIFF
--- a/lib/settings-panel.coffee
+++ b/lib/settings-panel.coffee
@@ -185,6 +185,7 @@ appendSetting = (namespace, name, value) ->
   if namespace is 'core'
     return if name is 'themes' # Handled in the Themes panel
     return if name is 'disabledPackages' # Handled in the Packages panel
+    return if name is 'customFileTypes'
 
   @div class: 'control-group', =>
     @div class: 'controls', =>


mnquintana:
Should we maybe add some test coverage for this?

maxbrunsfeld:
Personally, I think that the settings-view as a whole could use a lot better test coverage, but I don't really think it's in scope for *this* particular change.

For a change like this, I do think it's important to add tests that fit in with the project's existing tests. But for the change in question, the existing tests are [sparse](https://github.com/atom/settings-view/blob/master/spec/settings-panel-spec.coffee#L7). They don't make assertions about individual fields in Atom's config schema that are omitted from the UI, which I actually think is fine.

In the past, I've definitely accidentally broken things in settings-view, through changes to Atom core, without seeing any test failures. I'd love to help prevent this from happening in the future, but it'll be a bigger effort, and I don't think it needs to block this PR.

mnquintana:
Fair enough :+1:

maxbrunsfeld:
Thanks for :eyes: for our test coverage.

PR #674, opened 2015-10-03
Fix 100% CPU use on startup. Fixes #555 for me.

### LINKED ISSUE #555: Atom 1.0 will consume 100% CPU when managing packages
_From @lovio on June 26, 2015 6:7_

No matter installing or uninstalling packages, `Atom Helper` will comsume over 100% CPU.

_Copied from original issue: atom/atom#7438_

LegNeato:
This was causing a 100% cpu spike on launch with many packages (like Nuclide). Completely fixed my `Atom Helper` CPU issues on launch!

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -59,16 +59,17 @@ module.exports =
   consumeStatusBar: (statusBar) ->
     PackageManager = require './package-manager'
     packageManager = new PackageManager()
-    packageManager.getOutdated().then (packages) ->
-      if packages.length > 0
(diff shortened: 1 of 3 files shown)


LegNeato:
Note: I still get 100% CPU when viewing packages in the settings view.

thedaniel:
This looks good to me.

dorongutman:
Can you please release an new version of atom now that this is fixes ? It's impossible to work with atom like this!!

nathansobo:
@dorongutman We should have a new beta out this week.

PR #675, opened 2015-10-04
Add notes to Settings panel to point users to package settings

izuzak:
While we wait for https://github.com/atom/settings-view/issues/284 to be implemented, I think users would find it helpful if they were told that packages have settings of their own, while they're looking at the config settings in the Settings panel.

This pull requests adds two such notes: the first one tells users that packages have additional settings, and the other one tells users the some editor settings can be overridden on a per-language basis. Here's how that looks:

screen shot 2015-10-04 at 22 01 06

screen shot 2015-10-04 at 22 01 14

My non-native-English-speaking brain couldn't come up with better text for the notes -- I'd appreciate any suggestions about that.

What do you think, @atom/feedback? :thought_balloon:

### DIFF
--- a/lib/general-panel.coffee
+++ b/lib/general-panel.coffee
@@ -13,11 +13,11 @@ class GeneralPanel extends View
 
     @subPanels = [
       new SettingsPanel('core', note: '''
-        <div class="alert alert-info" id="core-settings-note">These are Atom's core settings which affect behavior unrelated to text editing. Individual packages might have additional config settings of their own. Check individual package settings by clicking its package card in the <a class="packages-open">Packages list</a>.</div>
+        <div class="text icon icon-question" id="core-settings-note" tabindex="-1">These are Atom's core settings which affect behavior unrelated to text editing. Individual packages might have additional config settings of their own. Check individual package settings by clicking its package card in the <a class="link packages-open">Packages list</a>.</div>
       ''')
 
       new SettingsPanel('editor', note: '''
-        <div class="alert alert-info" id="editor-settings-note">These config settings are related to text editing. Some of these settings can be overriden on a per-language basis. Check language settings by clicking the language's package card in the <a class="packages-open">Packages list</a>.</div>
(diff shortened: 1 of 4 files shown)


thomasjo:
@izuzak I think this looks really, really good. Great idea also! My only tiny little nitpick is about the word *selecting*; would perhaps *clicking on* be even more descriptive? I realize this level of nitpicking is likely ridiculous.

izuzak:
> My only tiny little nitpick is about the word selecting; would perhaps clicking on be even more descriptive?

I think it would -- thanks for mentioning it, @thomasjo :microscope:. Another idea in this direction: "Check individual package settings by _clicking its package card_ in the Packages list". :thought_balloon: ? :+1: / :-1:

thomasjo:
> "Check individual package settings by *clicking its package card* in the Packages list"

:heart:

lee-dohm:
:+1: This is a great idea and another :+1: on the wordsmithing.

simurai:
Agreed, this is a great addition. :+1:

One question: Is it too strong being wrapped in an `alert`? Could make it feel like a temporary message, more like the update alert:

screen shot 2015-10-06 at 11 44 43 am

Instead, it could be the same as the info `text` found on other pages:

screen shot 2015-10-06 at 11 58 02 am

that would look like:

screen shot 2015-10-06 at 12 02 59 pm

Might get less attention that way.

izuzak:
> Instead, it could be the same as the info text found on other pages:

@simurai I like that idea :sparkles: -- thanks for mentioning it. I pushed that change in 9e9feb8, but now I think there's not enough empty space between the note and the first checkbox (Audio Beep):

screen shot 2015-10-07 at 20 10 11

Do your :art: :eyes: see a way to improve that (I mean, I can style those elements manually myself, but wondering if there's already an existing style for such a thing or a better approach for adding such a note)? I think the empty space is larger on your screenshots in https://github.com/atom/settings-view/pull/675#issuecomment-145729951 so I'm hoping you do :smile:. Feel free to push directly to this branch :grin:

simurai:
Noticed that too.. in all the other places where `.text` gets used, the next element has some top margin. So I guess we can do it here too: https://github.com/atom/settings-view/commit/172e8964b5ff29d003ea33580d89f5b0ae3dd5b1

Ok, looks good.

izuzak:
You're a :star2:, @simurai -- thank you! Merging this now :boat:

Issue #645, opened 2015-08-26
"Settings" button is not shown for certain community packages


gsklee:
Please see https://github.com/sindresorhus/atom-editorconfig/issues/49 for further details, thanks.

PR #676, opened 2015-10-05
Show settings button on package cards for packages with grammars

izuzak:
This is an attempt to close https://github.com/atom/settings-view/issues/645. 

The problem with that package is that it doesn't have any config settings of it own, but it [does define a grammar](https://github.com/sindresorhus/atom-editorconfig/tree/master/grammars) so it has scoped settings as any other language package.

* 2af57d1 adds a failing spec to demonstrate the problem with the `packageHasSettings` method:

  ```
package manager
  ::packageHasSettings
    it returns true when the pacakge does not have config, but does define language grammars
      Expected false to be true. (spec/package-manager-spec.coffee:191:64)
```

* 9842dd1 modifies `packageHasSettings` so that it returns `true` if a grammar was loaded from that package, using a similar approach as in [`PackageGrammarsView.getPackageGrammars`](https://github.com/atom/settings-view/blob/cb09a2d80fcab4c91b3cf47e7d6fe02174a4b1c0/lib/package-grammars-view.coffee#L27-L29)

  ```
Finished in 21.308 seconds
134 tests, 505 assertions, 0 failures, 0 skipped
```

With this change, all packages which define a grammar will have a "Settings" button on the package card, which I think is good since those packages do have (scoped) settings.

cc @benogle (I noticed you made [some changes in this area](https://github.com/atom/settings-view/commit/e5069ac4617ee065a9eefb788da8da1496a77548) a few months back, @thedaniel @kevinsawicki for :thought_balloon: 

Before:

screen shot 2015-10-05 at 09 05 34
screen shot 2015-10-05 at 09 05 59

After:

screen shot 2015-10-05 at 09 04 15
screen shot 2015-10-05 at 09 04 31

### DIFF
--- a/lib/package-manager.coffee
+++ b/lib/package-manager.coffee
@@ -22,7 +22,7 @@ class PackageManager
     else
       false
 
-  packageHasSettings: (packageName) ->
+  packageHasSettings: _.memoize((packageName) ->
     grammars = atom.grammars.getGrammars() ? []
     for grammar in grammars when grammar.path
       return true if grammar.packageName is packageName
@@ -31,6 +31,7 @@ class PackageManager
     pack.activateConfig() if pack? and not atom.packages.isPackageActive(packageName)
     schema = atom.config.getSchema(packageName)
     schema? and (schema.type isnt 'any')
+  )
 
   runCommand: (args, callback) ->
     command = atom.packages.getApmPath()
--- a/lib/package-manager.coffee
+++ b/lib/package-manager.coffee
@@ -23,6 +23,10 @@ class PackageManager
       false
 
   packageHasSettings: (packageName) ->
+    grammars = atom.grammars.getGrammars() ? []
+    for grammar in grammars when grammar.path
+      return true if grammar.packageName is packageName
+
     pack = atom.packages.getLoadedPackage(packageName)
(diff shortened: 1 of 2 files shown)


thedaniel:
I think it looks good. How do you feel about memoizing the results of `packageHasSettings(package)`? I know it's just milliseconds but we're iterating over a bunch of stuff to get that true/false, and I'm not too worried about packages changing their "has settings" state between reloads of settings view.

izuzak:
Thanks for the :eyes:, @thedaniel :bow:

> How do you feel about memoizing the results of packageHasSettings(package)? I know it's just milliseconds but we're iterating over a bunch of stuff to get that true/false, and I'm not too worried about packages changing their "has settings" state between reloads of settings view.

I hadn't thought of that -- I think that's a neat idea. :bulb: I don't have a good intuition about problems which might be caused by this, but it seems fine to me as well. I added that in 1a29c2d (using `_.memoize`). 

Gonna merge this in a day or two if there's no other feedback. :ear:

tempusrname:
This no longer works. Disabling soft-wrap or running 'Editor: Toggle Soft Wrap' command does not disable line breaks.

PR #678, opened 2015-10-09
Hide config settings defined by language packages

izuzak:
This hides the following config settings from the Settings panel so that it doesn't confuse users: `commentStart`, `commentEnd`, `increaseIndentPattern`, `decreaseIndentPattern`, `foldEndPattern`. As mentioned [here](https://github.com/atom/atom/blob/1d5fa01dc46c52ab352622dcc6456fcbcfc7079c/src/config-schema.coffee#L100), these settings are used only by specific languages -- there are no global defaults.

@benogle @kevinsawicki Does this look okay to you? 

Also, I'm wondering if these config settings should be visible in language package settings? Currently, they are not visible, i.e. they're not listed [here](https://github.com/atom/settings-view/blob/b9567cd50448b91eb8f8f09f1dfe2efff156573b/lib/settings-panel.coffee#L16-L30). From the comment I linked to above it _seems_ as if they should be visible, but I'm not sure. Perhaps it's something that packages defined and shouldn't be touched by users? :thought_balloon: 

Before:

screen shot 2015-10-09 at 14 03 29

After:

screen shot 2015-10-09 at 14 03 04

### DIFF
--- a/lib/settings-panel.coffee
+++ b/lib/settings-panel.coffee
@@ -189,6 +189,9 @@ appendSetting = (namespace, name, value) ->
     return if name is 'disabledPackages' # Handled in the Packages panel
     return if name is 'customFileTypes'
 
+  if namespace is 'editor'
+    return if name in ['commentStart', 'commentEnd', 'increaseIndentPattern', 'decreaseIndentPattern', 'foldEndPattern'] # There's no global default for these, they are defined by language packages
+
   @div class: 'control-group', =>
     @div class: 'controls', =>
       schema = atom.config.getSchema("#{namespace}.#{name}")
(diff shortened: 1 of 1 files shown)


lib/settings-panel.coffee
```diff
@@ -189,6 +189,9 @@ appendSetting = (namespace, name, value) ->
     return if name is 'disabledPackages' # Handled in the Packages panel
     return if name is 'customFileTypes'
 
+  if namespace is 'editor'
+    return if name in ['commentStart', 'commentEnd', 'increaseIndentPattern', 'decreaseIndentPattern', 'foldEndPattern'] # There's no global default for these, they are defined by language packages
```
kevinsawicki: Super minor, but I think putting the comment above this line instead of at the end might make a bit more readable.

kevinsawicki:
I'm good with this, left one minor comment about comment placement. :+1:

lib/settings-panel.coffee
izuzak: > I think putting the comment above this line instead of at the end might make a bit more readable.

:+1: Changed in 536a1a0

Issue #638, opened 2015-08-20
Version of packages displayed as 'undefined' in full view


rugk:
![noversiondisplayed](https://cloud.githubusercontent.com/assets/11966684/9388820/24e8ed78-4768-11e5-8668-a3b1a2e6ecd7.gif)

This happens with every package I haven't installed.

Atom 1.0.7
Windows 7

Moved from: https://github.com/atom/atom/issues/8459

PR #679, opened 2015-10-09
Use metadata returned from atom.io API when creating package card in detail view

izuzak:
This fixes https://github.com/atom/settings-view/issues/638.

@thedaniel do you remember why you added this in https://github.com/atom/settings-view/pull/614/files#diff-078cae2173fad78e9ce04ce50aa4cf96R95 ? As far as I can tell, the response from the Atom.io API will include a `metadata` hash which has all the required properties, including the missing `version` (e.g. see https://atom.io/api/packages/pdf-view).

### DIFF
--- a/lib/package-detail-view.coffee
+++ b/lib/package-detail-view.coffee
@@ -93,9 +93,6 @@ class PackageDetailView extends ScrollView
         @showErrorMessage()
       else
         @pack = packageData
-        # TODO: this should match Package.loadMetadata from core, but this is
-        # an acceptable hacky workaround
-        @pack.metadata = _.clone(@pack)
         @completeInitialzation()
 
   showLoadingMessage: ->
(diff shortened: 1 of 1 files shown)


thedaniel:
I don't recall, unfortunately. I think it's possible that some other aspect of settings view expected or expects pack.metadata to contain some sort of top-level attribute like name or repo that was missing from the atom.io response, but I can't say for sure without digging in.

izuzak:
> I don't recall, unfortunately. I think it's possible that some other aspect of settings view expected or expects pack.metadata to contain some sort of top-level attribute like name or repo that was missing from the atom.io response, but I can't say for sure without digging in.

You were right, @thedaniel -- the `readme` property is not there, it's in the top-level object. I'm just going to use `_.extend` instead of `_.clone` (previous behavior) so that properties from both the top-level object and from the `metadata` object are used. 

Thanks!

izuzak:
:boat: :fingers-crossed:

PR #688, opened 2015-10-17
Show default in tooltip when present in settings - Resolves #268

### LINKED ISSUE #268: No indicator for what the default is for checkbox
screen shot 2014-10-22 at 8 30 32 am

`Non-word characters` and `Preferred line length` both clearly indicate that they are set to the default and what the default is. But none of the checkboxes give any indication of either whether they are set to the default nor what the default is.

I suggest adding a `Default: false/true` after checkboxes to give that visual indication.

johanderuijter:
Add a tooltip with the default value to the settings view.

Couple questions:
- At the moment i add the tooltips to the left. This means that the appear in roughly the same location for each setting, however, for input field with long defaults, this also means that the tooltip might fall outside the screen. Especially with the tree view 'off'.
- I don't know how to write tests to check if the tooltips are present when there is a default value.

Anyone got a couple pointers to help me finish this?

### DIFF
--- a/lib/settings-panel.coffee
+++ b/lib/settings-panel.coffee
@@ -39,6 +39,7 @@ class SettingsPanel extends CollapsibleSectionPanel
     @bindInputFields()
     @bindSelectFields()
     @bindEditors()
+    @bindTooltips()
     @handleEvents()
 
   dispose: ->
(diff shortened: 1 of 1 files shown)


johanderuijter:
In case anyone was wondering what it would look like.

screenshot 2015-11-08 21 37 42

jerone:
Simple and useful! :+1:

johanderuijter:
Updated the PR, resolved the conflict. Defaults are show for in te following manner:
atom settings view with default

Changed the placement to  `auto left` so should automatically move the the right when it doesn't fit there.

Also tried to add a test for the tooltip but still couldn't figure it out. Specifically, i dont know how to mimic the hover state.

lee-dohm:
Thanks for contributing and your persistence!

a7madgamal:
I like the idea, but it was a little confusing to me
`Default: true` gave me the impression that the default VALUE for this option is TRUE or enabled, not that the CURRENT setting is the default one.
Also, to have an overview of my settings, I need to hover over every single setting in all tabs :D

I would give a *very subtle* visual indication (different color bg, border, font...) for changed value, and only apply the tip with the default value in that case.

bonus: give me a way to revert the value for this field > 🥇

a7madgamal:
nonetheless: thanks for making atom a little more awesome <3

PR #690, opened 2015-10-25
Hide collapse icon when there are no themes/packages

### LINKED ISSUE #636: Hide collapse icon when there are no items shown
On Twitter: 
screen shot 2015-08-18 at 9 50 55 am

Was also thinking the whole title could be hidden, but it's probably still useful to know that a section exists even when showing `0/0`. Especially when doing a search. Maybe "Deprecated Packages" could be hidden since it's more a temporary section.

evjan:
If there are no packages or themes in a section, don't show it as collapsible and don't do anything if you click on it.

Packages before:
screen shot 2015-10-25 at 8 18 31 pm

Packages after:
screen shot 2015-10-25 at 8 18 05 pm

Themes before:
screen shot 2015-10-25 at 8 19 10 pm

Themes after:
screen shot 2015-10-25 at 8 17 49 pm

Fixes #636

### DIFF
--- a/lib/collapsible-section-panel.coffee
+++ b/lib/collapsible-section-panel.coffee
@@ -1,0 +1,33 @@
+{$$, TextEditorView, ScrollView} = require 'atom-space-pen-views'
+
+module.exports =
+class CollapsibleSectionPanel extends ScrollView
+  notHiddenCardsLength: (sectionElement) ->
+    sectionElement.find('.package-card:not(.hidden)').length
+
+  updateSectionCount: (headerElement, countElement, packageCount, totalCount) ->
+    if totalCount is undefined
+      countElement.text packageCount
+    else
+      countElement.text "#{packageCount}/#{totalCount}"
+
+    headerElement.addClass("has-items") if packageCount > 0
+
+  updateSectionCounts: ->
+    @resetSectionHasItems()
+
+    filterText = @filterEditor.getModel().getText()
+    if filterText is ''
+      @updateUnfilteredSectionCounts()
+    else
+      @updateFilteredSectionCounts()
+
+  handleEvents: ->
+    @on 'click', '.sub-section .has-items', (e) ->
+      e.currentTarget.parentNode.classList.toggle('collapsed')
+
+  resetCollapsibleSections: (headerSections) ->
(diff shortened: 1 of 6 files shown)


evjan:
There's a LOT of duplication between the packages and themes panels

thomasjo:
After a quick glance through the commits, this looks great. I'm a little bit concerned about the fact this PR also reduces the duplication. Don't get me wrong, that's a really great change, but that means that this PR does two things even though the title and description gives no indication about that fact. And the "deduplication" change does make it hard to separate those two changes, which makes it kinda tricky to revert either of the two.

I would have preferred to see the this split into two separate PRs personally, but I'm not sure it's worth the effort to do that at this point. Beyond this concern, this looks good to me. Great job.

/cc @atom/feedback

evjan:
@thomasjo Reverting the final dedup commit is easy enough to do and I'm happy to submit another PR for that if you want.

The reason I made it part of this PR is that the deduplication was mostly related to the new not-being-able-to-collapse functionality though, so it would probably make sense to revert both at the same time if need be. The boy scout rule of "always leave the campground cleaner than you found it" made me feel bad about sending off this PR without some cleaning up :)

thomasjo:
> The boy scout rule of "always leave the campground cleaner than you found it" made me feel bad about sending off this PR without some cleaning up :)

:heart: 

As for actually splitting it up, it would incur some extra work on your behalf that might not be worth it, so I'd say just leave it as it is. I'll leave the final decision up to you and/or anyone else with a strong opinion on the matter. Consider my sentiment more as a note for the future :bow:

evjan:
Duly noted, I'll leave it as is now and be mindful about that in the future. Unless someone else objects of cource

thedaniel:
I'm OK with including the refactor in this PR, the settings view needs all the help it can get ;)

thomasjo:
OK, cool. I'm gonna merge and :ship: this!

thomasjo:
@evjan Thanks for this! Awesome work :metal:

evjan:
Awesome, tusen takk for reviewing and providing feedback!

maxbrunsfeld:
Thanks @evjan! Nice work.

PR #691, opened 2015-10-31
Removed some themes and installed packages panel duplication

evjan:
Pulled up common behaviour to the super class


damieng:
Going to close this down as it hasn't been acted upon in the 18 months since it was opened.

PR #698, opened 2015-11-15
add optional sub-section heading if the settings group has an explicit title

dirk-thomas:
While package settings can have grouped options (see the below example) they are not visually separated. This patch adds a `sub-section-headline` if the group has an explicit `title` set. The visual grouping together with a reasonable title provides a better overview over the settings of a package if the number of options increases.

I didn't use `getSettingTitle` since that would introduce the new headlines where none have been before and I wanted to keep the changes to existing settings minimal.

    config:
      groupA:
        type: 'object'
        title: 'Options related to A'
        properties:
          optionA:
            type: 'string'
            default: 'default A'
          optionB:
            type: 'boolean'
            default: false
      groupB:
        type: 'object'
        title: 'Options related to B'
        properties:
          optionC:
            type: 'string'
            default: ''
          optionD:
            type: 'boolean'
            default: false

Since I haven't contributed to atom packages much yet I am not sure if the added html structure makes sense here or should not be changed in that way?

### DIFF
--- a/lib/settings-panel.coffee
+++ b/lib/settings-panel.coffee
@@ -292,6 +292,6 @@ appendObject = (namespace, name, value) ->
   title = getSettingTitle(keyPath, name)
   @div class: 'sub-section', =>
     @div class: 'sub-section-heading', title
-    @div class: 'section-body', =>
+    @div class: 'sub-section-body', =>
       for key in _.keys(value).sort()
         appendSetting.call(this, namespace, "#{name}.#{key}", value[key])
--- a/lib/settings-panel.coffee
+++ b/lib/settings-panel.coffee
@@ -290,5 +290,6 @@ appendObject = (namespace, name, value) ->
   title = getSettingTitle(keyPath, name)
   @div class: 'sub-section', =>
     @div class: 'sub-section-heading', title
-    for key in _.keys(value).sort()
-      appendSetting.call(this, namespace, "#{name}.#{key}", value[key])
+    @div class: 'section-body', =>
+      for key in _.keys(value).sort()
+        appendSetting.call(this, namespace, "#{name}.#{key}", value[key])
--- a/lib/settings-panel.coffee
+++ b/lib/settings-panel.coffee
@@ -286,6 +286,8 @@ appendArray = (namespace, name, value) ->
       @subview keyPath.replace(/\./g, ''), new TextEditorView(mini: true, attributes: {id: keyPath, type: 'array'})
 
 appendObject = (namespace, name, value) ->
+  return unless _.keys(value).length
+
   keyPath = "#{namespace}.#{name}"
   title = getSettingTitle(keyPath, name)
   @div class: 'sub-section', =>
--- a/lib/settings-panel.coffee
+++ b/lib/settings-panel.coffee
@@ -286,5 +286,9 @@ appendArray = (namespace, name, value) ->
       @subview keyPath.replace(/\./g, ''), new TextEditorView(mini: true, attributes: {id: keyPath, type: 'array'})
 
 appendObject = (namespace, name, value) ->
-  for key in _.keys(value).sort()
-    appendSetting.call(this, namespace, "#{name}.#{key}", value[key])
+  keyPath = "#{namespace}.#{name}"
+  title = getSettingTitle(keyPath, name)
+  @div class: 'sub-section', =>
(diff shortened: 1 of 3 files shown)


lee-dohm:
Can you give visual example of what this looks like? Also, can you include specs for the new behavior?

dirk-thomas:
Without the patch all four options are visually ungrouped:
before

With the patch the title of the group separates the different options visually:
after

This is only my second PR for Coffeescript. I haven't looked into spec files yet. I could try to but I am not sure how long it will take me.

dirk-thomas:
If I understand the `settings-panel-spec.coffee` spec file correctly it doesn't test any of the grouping functionality yet. So I am not sure what I could test specifically for this patch (without adding specs for a lot of existing functionality)?

simurai:
With the current settings, it seems only the "invisibles" have grouped options, there is no title, but it gets some padding.

screen shot 2015-11-17 at 12 23 44 pm

 Is the plan to group the current settings more. Like instead of 

- [ ] Auto Indent
- [ ] Auto Indent On Paste

have them grouped under "Indent":

- Indent
  - [ ] Auto Indent
  - [ ] Auto Indent On Paste

I think that would a good improvement.. :+1:

dirk-thomas:
The fact that groups without titles get the padding is a detail how the patch is implemented. It could be easily done that there is no extra padding if the group doesn't have a title. But imo the visual separation helps to structure the long list of settings.

Another option would be to always generate a missing group title from the group name (like it is done for each setting). But I don't know if existing group have reasonable enough names for that.

simurai:
> But imo the visual separation helps to structure the long list of settings.

Yes, I like that too.

dirk-thomas:
I have looked through the settings pages of all the packages installed in my Atom instance. I looks like that only very few of them use groups at all. Based on that I have updated the PR to always print the title for groups. If a group title isn't specified explicitly it is being derived from the group name (same logic as for each setting itself).

dirk-thomas:
@lee-dohm Can you please advice me what is necessary for this patch to be considered?

I am willing to add a spec if you can give me a hint what you expect to be tested. To me it looks like that the whole settings-panel isn't covered by any spec yet. Adding specs for all of the existing functionality seems to be not realistic.

lee-dohm:
You're not expected to add specs for the entire settings-view, just for the new functionality. Tests for:

* When there is a group
    1. The group title exists
    1. The elements of the group are displayed within the group
* When there is not a group
    1. No group title exists

dirk-thomas:
I have added specs for these cases in https://github.com/atom/settings-view/pull/698/files#diff-4bc58e9691f90d80eb8898b73290136d

dirk-thomas:
@lee-dohm Is there anything else I should update for this PR to be considered?

thedaniel:
The spacing between a section title and the settings is pretty cramped. Can we space it out a bit? cc @simurai for further UI feedback.

screen shot 2015-11-30 at 3 07 33 pm

dirk-thomas:
Since there was no feedback how the style should be applied I looked for reasonable existing classes and wrapped all the controls within a group into a tag with the `class=section-body` (see 15451fc87390cac46634c928687800bea5091e5c).

With that change there is more padding between the group headline and the block of controls: settings-view

dirk-thomas:
Now passing all the tests again.

simurai:
Looks good! :+1: from me.

dirk-thomas:
Can this package be released to make the new sub-section heading available to users?

thedaniel:
Sure thing. I'll bump the version today, which should put it in Atom 1.5.

dirk-thomas:
Thank you!

Arcanemagus:
I would imagine the reason nobody is using configuration groups is that it isn't documented at all, at least I couldn't find mention of it in the [Config API](https://atom.io/docs/api/v1.4.0/Config) or the [Config flight manual section](https://atom.io/docs/v1.4.0/behind-atom-configuration-api).

Is this functionality actually meant to be available yet, and if so should I file an issue on Atom to get it documented?

thedaniel:
An issue on atom/atom or a PR would be great!

dirk-thomas:
It actually is being mentioned briefly: https://atom.io/docs/api/v1.4.0/Config#object

But I thought it could use some improvement to be more discoverable: atom/atom#10552

PR #707, opened 2015-11-26
align search and error messages

mertkahyaoglu:
<b>Before:</b>
screen shot 2015-11-26 at 23 11 33
screen shot 2015-11-26 at 23 21 20

<b>After:</b>
screen shot 2015-11-26 at 23 23 28

### DIFF
--- a/styles/settings-view.less
+++ b/styles/settings-view.less
@@ -570,13 +570,12 @@
       }
     }
 
-    .search-message,
-    .featured-message {
-      margin: @component-padding @component-padding 0 @component-padding;
+    .search-message {
+      margin: @component-padding 0 0;
     }
 
     .error-message {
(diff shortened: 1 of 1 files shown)


styles/settings-view.less
```diff
@@ -570,13 +570,12 @@
       }
     }
 
-    .search-message,
-    .featured-message {
-      margin: @component-padding @component-padding 0 @component-padding;
+    .search-message {
```
lee-dohm: What about `.featured-message`?

simurai: Tried to find if `.featured-message` gets used, but doesn't seem so. Probably a left-over. I think it's ok to remove it.

simurai:
Looks good. @mertkahyaoglu thanks.

PR #715, opened 2015-12-23
Ask for confirmation before uninstalling a package

### LINKED ISSUE #699: Uninstall button of plugins should ask for confirmation
The "Uninstall" button of community plugins just remove, without asking for any confirmation.

schermata 2015-11-13 alle 09 23 51

This can lead to errors when the users click it by mistake (for example: his intention was to click "Settings" or "Disable")

Initially reported in atom/atom#9557

mgarciaisaia:
Ask the user for confirmation before uninstalling a package:

Asking for confirmation

Fixes #699

It probably needs a way to undo the operation - ie, if you clicked by mistake, the button will persist that way.

A timeout? How long?
Another 'Cancel' button? Uuggh :hankey:

### DIFF
--- a/lib/package-card.coffee
+++ b/lib/package-card.coffee
@@ -144,7 +144,15 @@ class PackageCard extends View
 
     @uninstallButton.on 'click', (event) =>
       event.stopPropagation()
-      @uninstall()
+      if @uninstallButton.hasClass('confirm-uninstall')
+        @uninstallButton.removeClass('confirm-uninstall')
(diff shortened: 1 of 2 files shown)


damieng:
Given it's so easy to undo the action - install an uninstalled package - do we need to confirm?  

Are these some negatives to clicking it by mistake such as losing settings or other destructive action?

damieng:
No response to query.

Issue #597, opened 2015-07-21
Order of a package's settings dialog


nikmartin:
Is there a way for a package author to group and order settings? A package like atom-beautify has a LOT of settings for each language, but the package author has no control over the order of grouping of the settings for that package. There are 15-20 settings options for every language atom-beautify supports, and they are sprinkled across the package settings dialog.

Glavin001:
I think having collapsible groups would be very helpful. Group all of the language options together and then collapse them until a User wants to expand the options for a specific language.

olmokramer:
You can set an `order` property on the items in your `config` object, and you can group by putting  those items in an object:

```coffee
exports.config = {
  group1: {
    type: 'object',
    order: 1,
    properties: {
      prop1: {
        type: 'integer',
        # ...
      },
      prop2: {
        # ...
      },
    },
  },
  group2: {
    type: 'object',
    order: 2,
    # ...
  },
};
```
Collapsible groups in the settings view would be a nice to have, though!

Glavin001:
Great, thanks @olmokramer ! It would be very nice to have these groups be collapsable. Is someone already working on that? Hopefully I can find time later this month to possibly contribute to that feature.

lee-dohm:
Since the original question about ordering and grouping was answered, I'm going to close this. Thanks for working on the PR @Glavin001 :grinning:

PR #736, opened 2016-02-21
Add collapsable section for option groups

Glavin001:
### About

For a package that has a lot of options, such as [`Atom Beautify`](https://github.com/Glavin001/atom-beautify), it appears to be very difficult for users to find the options they want.

I propose that the groups be collapsable.

#### Features

- [x] Groups of options are collapsable
  - Open by default
- [x] Force group to be collapsed

```javascript
{
  group1: {
    title: 'Group 1',
    type: 'object',
    collapsed: true
  }
}
```

- ~~"Collapse all sections" button (like the styleguide has)~~
  - ~~Only show "Collapse All" button when there are collapsable sub-sections~~

### Screenshots

Open `Group 1`:
image

Collapsed `Group 1`:
image

### Options used

```
  _group1: {
    title: 'Group 1',
    type: 'object',
    order: 1,
    properties: {
      prop1: {
        type: 'integer',
        title: 'Prop1',
        default: 1
      },
      prop2: {
        title: 'Prop2'
        default: 1
      },
    },
  },
  _group2: {
    title: 'Group 2',
    type: 'object',
    order: 2,
    properties: {
      prop1: {
        type: 'integer',
        title: 'Prop1',
        default: 1
      },
      prop2: {
        title: 'Prop2'
        default: 1
      },
    },
  }
```

/cc https://github.com/atom/settings-view/issues/597 https://github.com/Glavin001/atom-beautify/issues/620 https://github.com/Glavin001/atom-beautify/issues/405

### DIFF
--- a/lib/settings-panel.coffee
+++ b/lib/settings-panel.coffee
@@ -1,11 +1,12 @@
 {CompositeDisposable} = require 'atom'
 {$, $$, TextEditorView, View} = require 'atom-space-pen-views'
 _ = require 'underscore-plus'
+CollapsibleSectionPanel = require './collapsible-section-panel'
 
 {getSettingDescription} = require './rich-description'
 
 module.exports =
-class SettingsPanel extends View
+class SettingsPanel extends CollapsibleSectionPanel
   @content: ->
     @section class: 'section settings-panel'
 
@@ -39,6 +40,8 @@ class SettingsPanel extends View
     @bindInputFields()
     @bindSelectFields()
     @bindEditors()
+    @handleEvents()
+
 
   dispose: ->
     @disposables.dispose()
@@ -290,8 +293,10 @@ appendObject = (namespace, name, value) ->
 
   keyPath = "#{namespace}.#{name}"
   title = getSettingTitle(keyPath, name)
-  @div class: 'sub-section', =>
-    @div class: 'sub-section-heading', title
+
+  @section class: 'sub-section', =>
+    @h3 class: 'sub-section-heading has-items', =>
+      @text title
     @div class: 'sub-section-body', =>
       for key in _.keys(value).sort()
         appendSetting.call(this, namespace, "#{name}.#{key}", value[key])
--- a/lib/settings-panel.coffee
+++ b/lib/settings-panel.coffee
@@ -70,7 +70,7 @@ class SettingsPanel extends CollapsibleSectionPanel
             appendSetting.call(this, namespace, name, settings[name])
 
   sortSettings: (namespace, settings) ->
-    _.chain(settings).keys().sortBy((name) -> name).sortBy((name) -> atom.config.getSchema("#{namespace}.#{name}")?.order).value()
+    sortSettings(namespace, settings)
 
   bindInputFields: ->
     @find('input[id]').toArray().forEach (input) =>
@@ -186,6 +186,9 @@ isEditableArray = (array) ->
     return false unless _.isString(item)
   true
 
+sortSettings = (namespace, settings) ->
+  _.chain(settings).keys().sortBy((name) -> name).sortBy((name) -> atom.config.getSchema("#{namespace}.#{name}")?.order).value()
+
 appendSetting = (namespace, name, value) ->
   if namespace is 'core'
     return if name is 'themes' # Handled in the Themes panel
@@ -299,5 +302,6 @@ appendObject = (namespace, name, value) ->
     @h3 class: 'sub-section-heading has-items', =>
       @text title
     @div class: 'sub-section-body', =>
-      for key in _.keys(value).sort()
+      sortedSettings = sortSettings(keyPath, value)
+      for key in sortedSettings
         appendSetting.call(this, namespace, "#{name}.#{key}", value[key])
--- a/lib/settings-panel.coffee
+++ b/lib/settings-panel.coffee
@@ -293,8 +293,9 @@ appendObject = (namespace, name, value) ->
 
   keyPath = "#{namespace}.#{name}"
   title = getSettingTitle(keyPath, name)
-
-  @section class: 'sub-section', =>
+  schema = atom.config.getSchema(keyPath)
+  isCollapsed = schema.collapsed is true
+  @section class: "sub-section#{if isCollapsed then ' collapsed' else ''}", =>
     @h3 class: 'sub-section-heading has-items', =>
       @text title
     @div class: 'sub-section-body', =>
--- a/spec/settings-panel-spec.coffee
+++ b/spec/settings-panel-spec.coffee
@@ -67,6 +67,7 @@ describe "SettingsPanel", ->
                 default: false
           bazGroup:
             type: 'object'
+            collapsed: true
             properties:
               baz:
                 title: 'Baz'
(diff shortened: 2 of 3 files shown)


Glavin001:
Please let me know if there is anything I can do to make this merged and published quicker.

I'll take a look at the Travis CI failed tests once I am home.

Glavin001:
I have added tests and the ability to force a grouped setting to be collapsed by default.

Example usage:
    
 ```coffeescript
    barGroup:   # Expanded by default
        type: 'object'
        title: 'Bar group'
        properties:
          bar:
            title: 'Bar'
            description: 'The bar setting'
            type: 'boolean'
            default: false
    bazGroup:   # Collapsed by default
        type: 'object'
        collapsed: true # field to force collapse a grouped setting
        properties:
          baz:
            title: 'Baz'
            description: 'The baz setting'
            type: 'boolean'
            default: false
```

**Comment by thedaniel** [comment]
@simurai do you have feelings about this

**Comment by jerone** [comment]
With many groups, what about a "Collapse/expand all sections" (like the styleguide has)...

**Comment by Glavin001** [comment]
> With many groups, what about a "Collapse/expand all sections" (like the styleguide has)...

I'll look into that, too. Good idea. Thanks.

**Comment by Glavin001** [comment]
## Collapse All button

### Before: Expanded sections
image

### After: Collapsed all
image

**Comment by Glavin001** [comment]
The failing test on Travis CI is not influenced by my changes and only happens intermittently.

**Comment by Glavin001** [comment]
Please restart the Travis CI build to see that the tests do pass.

Someone should probably fix whatever is causing the timeout error:
```
PackageDetailView
  it displays the snippets registered by the package
    timeout: timed out after 60000 msec waiting for something to happen
```

**Comment by simurai** [comment]
I think it's great. Maybe not relevant for most packages, but looking at `atom-beautify`, I can see the need for it. :laughing:

Should the "Collapse All Sections" button be an "Expand All Sections" button instead. For `atom-beautify` it probably makes more sense to collapse all sections by default, except maybe "General". Then you just open the ones you're interested in. vs. "ohh this is overwhelming, let's first collapse everything".

In that case, the button could be disabled, once everything is expanded.

Some packages might start to show the most essential settings ungrouped and then a collapsed "Advanced settings" group that is a bit more hidden for less common settings. I think that would be a nice  pattern.

**Comment by Glavin001** [comment]
I have work and a midterm to study for so it is unlikely I will be making any changes until at the earliest Thursday night.

That gives everyone some time to chime in, discuss, and settle on *exactly* what they want so that this can be *reviewed and merged ASAP*.

Questions to answer/decide:
- Should it be a "Collapse All" or "Expand All" button?
- Is the default to have groups be collapsed or expanded? (currently default is expanded and you use `collapsed:true` to force group being collapsed by default)
- Any implementation changes?
- Any User Interface changes?

Thank you.

**Comment by Glavin001** [comment]
Any review of this pull request? Is there anything that needs to be changed before this can be merged? Thank you.

**Comment by Glavin001** [comment]
Any status on review of this pull request? I know there are at least many Atom Beautify users who are excited for this. Thanks in advance for merging :stuck_out_tongue_winking_eye:.

**Comment by thedaniel** [comment]
- Regardless of the default, in your screenshot the "Collapse all" button still says "Collapse" even after all the sections are collapsed. It should be "Expand" in that state.
- What does a mix of grouped and ungrouped settings render like in this state? (Is that even possible with our config schema?)

**Comment by Glavin001** [comment]
> Regardless of the default, in your screenshot the "Collapse all" button still says "Collapse" even after all the sections are collapsed. It should be "Expand" in that state.

This is the same as how it is implemented in the Styleguide. It also shows "Collapse All Sections" when all sections are already collapsed:

image

Clicking "Collapse All Sections" toggles the individual `collapse` classes on each of the `.sub-section`s.
See https://github.com/Glavin001/settings-view/blob/master/lib/settings-panel.coffee#L84 
If we want it to show `Expand` then we would have to observe the changes to all `.sub-section` elements and check if each of them are collapsed.

I think that while the "Collapse" / "Expand" functionality is a good idea, since other views, such as Stlyeguide, already have a simpler "Collapse All Sections" functionality that this should be a separate Issue. If we want "Collapse" / "Expand" then that should be another Pull Request specifically targeting the overall functionality of Collapsing and add support for Expanding.

If this is still a large concern, then I would be forced to suggest that we remove the "Collapse All Sections" button in this Pull Request all together and someone can commit the "Collapse" / "Expand" functionality on their own. This Pull Request is specifically targeting the collapsable sub-sections for within Settings views. The "Collapse All Sections" button was a nice-to-have that I thought would help create excitement and encourage this Pull Request to be closed quicker. However, if that is not the case, then I will promptly remove it in hopes that the core feature, collapsable settings, will be merged promptly.

> What does a mix of grouped and ungrouped settings render like in this state? (Is that even possible with our config schema?)

See https://github.com/atom/settings-view/pull/736#issuecomment-187699892 for screenshot.

**Comment by Glavin001** [comment]
Please let me know if I should make a new Pull Request with only the `Collapse` buttons for each sub-section -- the core purpose of this Pull Request -- and remove the seemingly controversial `Collapse All Sections` button that is feature creeping for this Pull Request. Thank you.

**Comment by prettydiff** [comment]
+1

**Comment by simurai** [comment]
The Styleguide doesn't necessarily have to serve as the blueprint for this. I think over time the list became too long and then a button got added to collapse everything so you don't have to scroll too much. But personally I would prefer if all the sections in the Styleguide are already collapsed initially.

> If this is still a large concern, then I would be forced to suggest that we remove the "Collapse All Sections" button in this Pull Request

:+1:  Ya, maybe a good idea to find agreement in a separate PR.

**Comment by Glavin001** [comment]
Given this Pull Request has taken this long to review, I have simplified it to hopefully make it more appealing so it can be merged sooner.

This Pull Request now only adds the following functionality:
- [x] Collapsable groups
- [x] Force group be start collapsed with option schema
- [x] Options within group are sorted
- [x] New specs for this functionality (and Travis CI says they pass!)

Thank goodness for atomic commits and rebasing.

**Comment by simurai** [comment]
I tested this in the "Editor Settings" to toggle the Invisibles group and it works as expected. :+1:

Maybe some final :eyes: on the code would be great.

**Comment by Glavin001** [comment]
Is there anything else I can do for this issue to be reviewed by those who have merge permission? I thought adding specs passing Travis CI would help expedite this process. Please let me know how else I can help this move along so Atom and Atom Beautify users can benefit. Thank you!

File: lib/settings-panel.coffee [hunk 1]
```diff
@@ -67,7 +70,7 @@ class SettingsPanel extends View
             appendSetting.call(this, namespace, name, settings[name])
 
   sortSettings: (namespace, settings) ->
-    _.chain(settings).keys().sortBy((name) -> name).sortBy((name) -> atom.config.getSchema("#{namespace}.#{name}")?.order).value()
+    sortSettings(namespace, settings)
```
thedaniel: Just curious, why did you move this out into a separate function since it's only used here?

thedaniel:
I left one comment otherwise I think it is good to merge.

lib/settings-panel.coffee
Glavin001: I use `sortSettings` in two places:
- https://github.com/Glavin001/settings-view/blob/master/lib/settings-panel.coffee#L62
- https://github.com/Glavin001/settings-view/blob/master/lib/settings-panel.coffee#L305

I moved the `sortSettings` functionality out since I use `sortSettings` outside of the `SettingsPanel` class at https://github.com/Glavin001/settings-view/blob/master/lib/settings-panel.coffee#L294-L307

This function could be removed and code changed from `@sortSettings` to simply calling `sortSettings`: https://github.com/Glavin001/settings-view/blob/master/lib/settings-panel.coffee#L72-L73
```
sortSettings: (namespace, settings) ->
    sortSettings(namespace, settings)
```

However, now I look at https://github.com/atom/settings-view/blob/e9ac7219a65756e48707f9aa720ea540b2846881/spec/settings-panel-spec.coffee#L42-L50 and see there are spec that use `settingsPanel.sortSettings` function.

Glavin001:
Could someone restart the Travis CI build? Something appears to have gone wrong. It does not look like it ever reached running the specs.

thedaniel:
OK, I really appreciate your patience! This should help with the pathological case of config size without changing the experience of the 99% case. It's just in time to get into the 1.7 beta this week.

Glavin001:
Awesome!!! Thank you very much for reviewing and merging. Happy to contribute to Atom.

PR #748, opened 2016-03-13
Add defaults on focus

### LINKED ISSUE #243: Defaults should be editable
Just installed atom. Being a python developer I wanted to hide *.pyc from tree view, but I also wanted to maintain the default ignored names. The problem is ".git, .hg, .svn, .DS_Store, Thumbs.db" dissapears when the textbox is clicked, and the user has to type it all again

fvantienen:
Add defaults on focus and remove them when input field contains the default.
This should fix #243.

### DIFF
--- a/lib/settings-panel.coffee
+++ b/lib/settings-panel.coffee
@@ -106,7 +106,9 @@ class SettingsPanel extends View
   isDefault: (name) ->
     params = {sources: [atom.config.getUserConfigPath()]}
     params.scope = [@options.scopeName] if @options.scopeName?
-    not atom.config.get(name, params)?
+    defaultValue = @getDefault(name)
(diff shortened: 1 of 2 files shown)


amytruong:
Took a quick look and tested per @lee-dohm's request (windows and mac) and it looks good on my end so far. :tada:

joshaber:
Looks good, thanks @fvantienen!

stefansundin:
Hey all, I am pretty sure this broke setting "0" for integer values.

In `isDefault`:

```coffee
not value or defaultValue is value
```

The `not value` part is true if `value=0`.

Apologies if something else did it, but it's broken alright. Let me know if you want me to open a new issue for it.

PR #755, opened 2016-03-22
Keep Syntax or UI when parts of the actual theme name

hartator:
Avoids removing Syntax or UI when they are an actual part of the theme name.

For example: https://atom.io/themes/no-syntax-highlighting-syntax

### DIFF
--- a/lib/themes-panel.coffee
+++ b/lib/themes-panel.coffee
@@ -270,7 +270,7 @@ class ThemesPanel extends CollapsibleSectionPanel
 
   # Get a human readable title for the given theme name.
   getThemeTitle: (themeName='') ->
-    title = themeName.replace(/-(ui|syntax)/g, '').replace(/-theme$/g, '')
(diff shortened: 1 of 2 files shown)


50Wliu:
Thanks for catching this @hartator :smile:.  Can you please add a spec for this to make sure it doesn't regress in the future?

hartator:
Sure, :)

I've added specs to test the way theme titles are displayed as there were none. I was also able to fully test half of it. 

For some reasons that I didn't found out, on my local Atom.io install, this line timeouts https://github.com/atom/settings-view/blob/master/spec/themes-panel-spec.coffee#L91 and all the related specs don't run. Consequently, I was able to try out the title display of the installed themes, but not mock a new one to try out the regression.

damieng:
There's a broken test here that needs fixing before we can merge.

damieng:
No response to request to fix test, closing.

Issue #384, opened 2015-02-17
Feature request: Allow sorting packages


boustanihani:
Could you please allow **sorting** packages (in the settings-view) by the **number of downloads** and the **number of stars** ?

bastilian:
I think this feature would make finding the right package so much easier and the change looks to be trivial to add. I mocked something up how I was thinking of adding the feature to the interface, the style needs fine tuning:

![screen shot 2016-04-05 at 08 35 10](https://cloud.githubusercontent.com/assets/7757/14273306/eaa29f1e-fb09-11e5-84e7-5e840a73bb9f.png)

Another feature I was thinking about, was to allow star a package in the install view and possibly a "my starred packages" filter as it may be an incentive to increase star counts and give an easy way to install all your favourite packages.

Let me know what you think and I'll add the sorting asap. It really bugs me too. ;)

bastilian:
I started to implement this in #761

Issue #513, opened 2015-05-20
Package Search doesn't find the most relevant package


bronson:
Search for `file-icons` on the Install tab.

Any idea why `file-type-icons` sorts above `file-icons`?  file-type-icons is both more obscure and doesn't match the search string exactly.

![screen shot 2015-05-19 at 3 51 08 pm](https://cloud.githubusercontent.com/assets/1776/7735836/a72ec184-fef4-11e4-9014-6fdd0550245a.png)

This may be related to https://github.com/atom/settings-view/issues/332

Atom 0.200.0

xPaw:
Yeah I keep noticing this too. It usually finds some obscure packages first above what I want.

calebmeyer:
Ditto (but worse) for the `react` package. It's way down the list for me in the atom-beta, and didn't even show up for a coworker on stable.

We should sort exact matches to the top.

pcarn:
Just came across the same thing with the react package. It's an exact match on the name, and many more downloads than the others, why isn't it shown first?

Ben3eeE:
A PR has now been merged which addresses this. See https://github.com/atom/settings-view/pull/761#issuecomment-228257289.

Issue #141, opened 2014-06-13
Add stars support


kevinsawicki:
- star packages
- unstar packages
- view your starred packages

PR #761, opened 2016-04-05
Sort packages search

bastilian:
This implements sorting of packages based on #384.

Currently it sorts package and theme search results based on downloads count. 

screen shot 2016-04-05 at 15 07 23

screen shot 2016-04-05 at 15 06 51

### DIFF
--- a/lib/install-panel.coffee
+++ b/lib/install-panel.coffee
@@ -159,6 +159,8 @@ class InstallPanel extends ScrollView
 
     opts = {}
     opts[@searchType] = true
+    opts['sortBy'] = "downloads"
+
     @packageManager.search(query, opts)
       .then (packages=[]) =>
         if packages.length is 0
--- a/lib/install-panel.coffee
+++ b/lib/install-panel.coffee
@@ -163,18 +163,35 @@ class InstallPanel extends ScrollView
 
     @packageManager.search(query, opts)
       .then (packages=[]) =>
-        if packages.length is 0
-          @searchMessage.text("No #{@searchType.replace(/s$/, '')} results for \u201C#{query}\u201D").show()
-        else
-          @searchMessage.hide()
+        @resultsContainer.empty()
+        packages
+      .then (packages=[]) =>
+        @searchMessage.hide()
+        @showNoResultMessage if packages.length is 0
+        packages
+      .then (packages=[]) =>
+        @highlightExactMatch(@resultsContainer, query, packages)
+      .then (packages=[]) =>
         @addPackageViews(@resultsContainer, packages)
       .catch (error) =>
         @searchMessage.hide()
         @searchErrors.append(new ErrorView(@packageManager, error))
 
-  addPackageViews: (container, packages) ->
-    container.empty()
+  showNoResultMessage: ->
+    @searchMessage.text("No #{@searchType.replace(/s$/, '')} results for \u201C#{query}\u201D").show()
 
+  highlightExactMatch: (container, query, packages) ->
+    exactMatch = _.filter packages, (pkg) ->
+      pkg.name is query
+
+    if exactMatch.length > 0
+      packageCard = @getPackageCardView(exactMatch[0])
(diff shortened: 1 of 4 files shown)


lee-dohm:
@simurai How do you feel about moving the version out of the metadata area and putting it next to the package name?

simurai:
> How do you feel about moving the version out of the metadata area and putting it next to the package name?

I kinda like it. It ties the name and version more together. Maybe it could be toned down a bit or smaller font-size so it stands out a bit more from the name?

One issue with sorting by downloads is that an exact match (but not as popular) might get under lots of other popular packages. Like if you copy&pasted the name and want to install exactly that package, it would be hard to find.

screen shot 2016-06-23 at 4 00 56 pm

Not sure if it's possible to sort with multiple criterias?

Also.. if we add the stars count, it would be nice if you could star packages too. See https://github.com/atom/settings-view/issues/141

bastilian:
Glad you like it @simurai. I'll make the version number to be smaller than the name.

Regarding the search results, I think putting an exact match as the first would make total sense and then have the following results sorted by stars. maybe even with a little bit of a gap between so the first and exact matching one sticks out even more.

Adding star and unstar to the package card shouldn't be a hard thing to do, as well as adding another list to the "Packages" pane to list your starred packages. For the "Packages" pane I would however suggest to default to collapsed lists, maybe revamp it even a little bit in the long run.

I'll style the version number better and implement the result listing.

Let me know what you think about how to add the Star-Features.

bastilian:
hm... I think past me had a thought... I implemented it in a way that the sort can be an option to the query, what do you think about adding two buttons like packages & themes, but smaller, below them to choose the sorting and default to downloads?

simurai:
I see, yeah. Adding the stars count would make more sense if it's also possible to sort by stars.

> list your starred packages

Would also be nice, but better to keep PRs small.

bastilian:
> Would also be nice, but better to keep PRs small.

Totally agree. I'll open a PR based on #141 and leave the stars out of here.

I'll make the changes for the search result and update this PR.

bastilian:
Oh, and we can add the sorting by stars in the other PR as well.

bastilian:
Tada! 🎉 !

screen shot 2016-06-23 at 11 37 33

* the version number is now `.8em` of the package name
* Star and downloads count are now configurable on PackageCard's (defaults to showing the downloads count for now.
* An exact match is now extracted and put in front of all other results.

50Wliu:
> An exact match is now extracted and put in front of all other results.

:heart_eyes:

Ben3eeE:
> An exact match is now extracted and put in front of all other results.

Does this apply to https://github.com/atom/settings-view/issues/513?

Same issue exists on the apm repo for the react package.

bastilian:
@Ben3eeE yes, this will address #513 as well.

I just tested with the `react` package, but noticed a bug that it appears twice in the result.
Will look into this.

bastilian:
Ok, that bug is now fixed. `react` and all other exact matches show up first now. :)

On a side note: I think this should be seen as a workaround, the actual fix should be applied to where the search is actually performed to have the same result in atom, apm and on atom.io. But this is a good testing ground to figure out how the search in general could work, what it should show as results and how they should be ordered.

lee-dohm:
Looks good. Thanks @bastilian! :bow:

lee-dohm:
@bastilian Is this still `[WIP]`? I tested it and it looks good ... but I didn't want to merge it unless you were done with it :grinning:

Ben3eeE:
I tested this as well. Everything seems to be working for me. Great work @bastilian :+1:

bastilian:
perfect. :) and it's not [WIP] anymore. :D

lee-dohm:
Thanks for this @bastilian :grinning:

bastilian:
YAY! Thank you for merging @lee-dohm!

ricksterhd123:
This no longer applies today, what happened?

PR #763, opened 2016-04-06
Show settings after install

### LINKED ISSUE #650: Settings button not available in Install view, post-package-install
Settings button not available in Install view, post-package-install.

example

After installing, we have to switch over to Packages view and search for the newly installed package to access the package's settings.

example2

It would be nice to show the settings button right in the Install view after a package is finished installing (if the package has settings to show of course) :)

bastilian:
This PR fixes #650 by not removing the settings button and show after a package has been installed that has settings.

For this I had to remove `_.memoize` for `PackageManager#packageHasSettings` as it returned a wrong result, this may effect rendering of longer package lists, but so far I have not noticed any negative effect.

### DIFF
--- a/lib/package-card.coffee
+++ b/lib/package-card.coffee
@@ -193,6 +193,8 @@ class PackageCard extends View
     @versionValue.text(@installablePack?.version ? @pack.version)
     if @pack.apmInstallSource?.type is 'git'
       @downloadCount.text @pack.apmInstallSource.sha.substr(0, 8)
+    if @hasSettings()
+      @settingsButton.show()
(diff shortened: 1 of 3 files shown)


lee-dohm:
There is a bug in this implementation. When I'm on a package's settings page, the settings button is displayed:

screen_shot_2016-06-21_at_8_43_26_am

bastilian:
@lee-dohm I've added a spec to verify this and fixed it. The settings button in now shown after install and not shown on the settings page of a package.

lee-dohm:
Thanks very much @bastilian!

PR #765, opened 2016-04-07
Set active panel with name and options

### LINKED ISSUE #626: Package detail page is not serialized
If you view a package's details and reload the editor, you will be shown the root settings view rather than the package you were looking at.

cf https://github.com/atom/settings-view/pull/614#issuecomment-129748322

bastilian:
This is to allow InstalledPackagesPanel to initialise with a previously viewed package and fixes #626.

I renamed `activePanelName` to reflect that it is now a panel object not just the name.


lee-dohm:
Thanks for contributing!

Issue #532, opened 2015-06-09
Package cards in Updates panel don't change state when updates are done installing 


mnquintana:
Just noticed this as of Atom 0.208.0-c06d62e:

1. Click the `Update` button in an available update package card in the Updates panel
2. Wait a sufficiently long time so you know it should've been installed (~5 mins with an SSD, for example)
3. Package card will still say "Updating..."
4. Click another panel in Settings View
5. Close Settings View
6. Reopen Settings View
7. Click the Updates panel
8. Assuming there was only 1 update, it will report that all packages are up-to-date.

mnquintana:
I also reproduced this with 0.207.0 (both on OS X 10.10.3). The two packages I updated were atom-beautify and git-plus, in case that turns out to be relevant.

mnquintana:
I've consistently reproduced this with atom-beautify on every recent update; but, for comparison, the button state changed correctly with minimap.

mertzt89:
This has also happened to me. In my case atom-beautify would appear to never finish updating even though it has. When re-opening the update panel it says everything is up to date.

mnquintana:
@benogle Do you think this was fixed by https://github.com/atom/settings-view/pull/572 by any chance?

benogle:
Maybe. I havent seen this specific issue, though. Was the package disabled? Or in some other state? There were a bunch of issues with disabled packages that I fixed today.

mnquintana:
Nope, it was in a normal state, but I experienced it very consistently with atom-beautify. I'll see if I get it again now with the changes in master.

mnquintana:
Yup, still an issue with atom-beautify :confused:

mnquintana:
I think this happens if a package doesn't provide a `deactivate` method in the version that's being replaced in the update - all of the packages I've been able to reproduce this with have no `deactivate` method (including atom-beautify).

jtokoph:
Figured I should note that the deactivation hook didn't fix this for me in git-plus (https://github.com/akonwi/git-plus/commit/9643e37cdf8a8b8b557f43a3c6b53399e610ab13). The change was introduced in 5.4.3 and I'm updating from 5.4.4 to 5.4.6.

![screen shot 2015-08-24 at 3 56 18 pm](https://cloud.githubusercontent.com/assets/346225/9454706/f87e6650-4a78-11e5-8208-2e1db81dbf68.png)

fuzzy76:
still happening

m1r4ge:
[...]

PR #771, opened 2016-04-20
Wait for updated package to load and then try to activate

jtokoph:
This is a fix for issue #532 - Package cards in Updates panel don't change state when updates are done installing.

Before this commit, updating a package that had activation commands would cause the package card to display the loading state forever.

The cause of the issue was that the update process was waiting for the updated package to activate. This was a problem because calling `activatePackage` will only immediately activate a package that doesn't have any activationCommands.

The proposed fix is to wait for the package to load and then attempt to activate the package without waiting.

### DIFF
--- a/lib/package-manager.coffee
+++ b/lib/package-manager.coffee
@@ -274,12 +274,13 @@ class PackageManager
     exit = (code, stdout, stderr) =>
       if code is 0
         @clearOutdatedCache()
-        activation = if activateOnSuccess
-          atom.packages.activatePackage(name)
-        else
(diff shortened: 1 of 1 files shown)


alexchandel:
Is this mergeable?

Stanzilla:
@lee-dohm @damieng any chance we could get this in before 1.9 release?

lee-dohm:
@Stanzilla If it gets merged now, it would be on track for v1.10. Taking a look at it now though.

Stanzilla:
Ah ok, on Beta anyway so all good :)

PR #773, opened 2016-04-21
Allow to search keybindings using multi-keyword query

octref:
Edit: Sorry for not having included more description about the changes in the PR.

## Problem
Currently, for a keybinding
```
Keystroke: alt-cmd-left
Command: pane:show-previous-item
Source: Core
Selector: body
```
A search like `pane show`, `pane cmd`, `Core`, `pane ` (whitespace at end) would all fail to hit it.

The problem is, for each keybinding, `searchString` is created:
`searchString = "#{selector}#{keystrokes}#{command}#{source}".toLowerCase()`
Then the input is tested to see if it's a substring of it.
This doesn't make sense, as common use cases like the ones I listed above aren't possible with this approach.

Two other small issues:
- Whitespace at the end of input is not trimmed
- `#{source}` refers to the keybinding file location like `/Applications/Atom.app/Contents/Resources/app.asar/keymaps/darwin.json`. On the screen it's displayed as `Core` after processing it with `KeybindingsPanel.determineSource`. It doesn't make sense to use the raw source in the `searchString`.

This PR fixes the two small issues, and breaks down input to keywords and test if each keyword is a substring in the `searchString`.
Sorry for my original misuse of the word "fuzzy match". As @jeancroy pointed out, this has nothing to do with fuzzy matching. I've edited the title to reflect it.

---
## Effect

image

Haven't been following Atom for a while but it's got much better since I last checked.
Thinking about doing some more contributions :D

### DIFF
--- a/lib/keybindings-panel.coffee
+++ b/lib/keybindings-panel.coffee
@@ -73,10 +73,13 @@ class KeybindingsPanel extends ScrollView
     @keybindingRows.empty()
     for keyBinding in keyBindings
       {selector, keystrokes, command, source} = keyBinding
+      source = KeybindingsPanel.determineSource(source)
       searchString = "#{selector}#{keystrokes}#{command}#{source}".toLowerCase()
       continue unless searchString
 
-      if /^\s*$/.test(filterString) or searchString.indexOf(filterString?.toLowerCase()) isnt -1
+      keywords = filterString.trim().toLowerCase().split(' ')
+      containKeyword = (keyword) -> return /^\s*$/.test(keyword) or searchString.indexOf(keyword) isnt -1
+      if keywords.every containKeyword
         @appendKeyBinding(keyBinding)
 
   appendKeyBindings: (keyBindings) ->
--- a/lib/keybindings-panel.coffee
+++ b/lib/keybindings-panel.coffee
@@ -78,8 +78,7 @@ class KeybindingsPanel extends ScrollView
       continue unless searchString
 
       keywords = filterString.trim().toLowerCase().split(' ')
-      containKeyword = (keyword) -> return /^\s*$/.test(keyword) or searchString.indexOf(keyword) isnt -1
-      if keywords.every containKeyword
+      if keywords.every((keyword) -> searchString.indexOf(keyword) isnt -1)
         @appendKeyBinding(keyBinding)
 
   appendKeyBindings: (keyBindings) ->
--- a/spec/keybindings-panel-spec.coffee
+++ b/spec/keybindings-panel-spec.coffee
@@ -89,3 +89,14 @@ describe "KeybindingsPanel", ->
       expect(row.find('.command').text()).toBe 'window:toggle-full-screen'
       expect(row.find('.source').text()).toBe 'Core'
       expect(row.find('.selector').text()).toBe 'body'
+
+    it "perform a fuzzy match for each keyword", ->
+      panel.filterKeyBindings keyBindings, 'core ctrl-a'
+
+      expect(panel.keybindingRows.children().length).toBe 1
+
+      row = panel.keybindingRows.children(':first')
+      expect(row.find('.keystroke').text()).toBe 'ctrl-a'
+      expect(row.find('.command').text()).toBe 'core:select-all'
+      expect(row.find('.source').text()).toBe 'Core'
+      expect(row.find('.selector').text()).toBe '.editor, .platform-test'


50Wliu:
In other parts of Atom, fuzzy matching is provided by [`fuzzaldrin-plus`](https://github.com/jeancroy/fuzzaldrin-plus) - perhaps you might want to take a look at that and see if you can integrate it here?

octref:
@50Wliu 
I feel that would be an overkill. The current problem is that if I enter `pane show`, nothing returns.
It doesn't handle whitespace or `Source` properly now, too.

BTW, I don't know why the CI is failing. I ran it on my machine and it passed.
The failing test https://github.com/atom/settings-view/blob/master/spec/installed-package-view-spec.coffee#L34-L62 don't seem to be related to what I changed.
Is it because I didn't emit some sort of event?

octref:
Just took a look at https://github.com/jeancroy/fuzzaldrin-plus, it does **character-wise fuzzy matching**, which would be computationally expensive, and less effective than word-wise fuzzy matching.

50Wliu:
Don't worry about the specs - I believe the failing one is a bit flaky right now.

Your explanation for wordwise fuzzy searching vs character-wise seems pretty reasonable, but I'm no expert in this field :laughing:.  Going to /cc @jeancroy for his input on this just in case.

lib/keybindings-panel.coffee
```diff
@@ -73,10 +73,13 @@ class KeybindingsPanel extends ScrollView
     @keybindingRows.empty()
     for keyBinding in keyBindings
       {selector, keystrokes, command, source} = keyBinding
+      source = KeybindingsPanel.determineSource(source)
```
50Wliu: Why was this added?

lib/keybindings-panel.coffee
```diff
@@ -73,10 +73,13 @@ class KeybindingsPanel extends ScrollView
...
       searchString = "#{selector}#{keystrokes}#{command}#{source}".toLowerCase()
       continue unless searchString
 
-      if /^\s*$/.test(filterString) or searchString.indexOf(filterString?.toLowerCase()) isnt -1
+      keywords = filterString.trim().toLowerCase().split(' ')
+      containKeyword = (k) -> return /^\s*$/.test(k) or searchString.indexOf(k) isnt -1
```
50Wliu: I think you can be more explicit here by renaming `k` to `keyword`

jeancroy:
@octref  yes it's more expensive. However it's still somewhat into the interactive time.

Looking at the source how you do "word fuzzy matching", basically I find nothing fuzzy. 
It's an indexOf, so a normal text search. Maybe tokenized so it'll allow out of order words.

fuzzaldrin-plus as well as normal fuzzaldrin would have problem dealing with this.
It don't really deal with different word belonging to different field of the filtered object.
Actually it don't support filtering by more than one field

lib/keybindings-panel.coffee
octref: The original source is something like:
```
/Users/octref/.atom/packages/vim-mode-plus/keymaps/vim-mode-plus.cson
/Applications/Atom.app/Contents/Resources/app.asar/keymaps/darwin.json
```
This makes them
```
Vim Mode Plus
Core
```

Which is more what user would search for, I think.

octref:
@jeancroy
Thanks for your input. I don't really know the proper name of this kind of search :p Maybe you can tell me so I can make the commit message more accurate.

lib/keybindings-panel.coffee
octref: Done!

jeancroy:
Yeah. Sorry to be pointy on this. It's just that if we say some field support fuzzy matching. User will expect it'll work like fuzzy-finder or command-pallette or autocomplete.

I'm not sure how this should be called. Maybe something like multi-keyword search ?
Actually I'm not sure what has been fixed. Do you have a before/after example of something that now work but didn't before ?

octref:
@jeancroy

Before this, for something like
```
Keystroke: alt-cmd-left
Command: show-previous-item
Source: Core
Selector: body
```
`pane show`
`pane cmd`
`pane core`
`alt cmd`
`pane ` (whitespace at end)
All wouldn't hit.
After the fix all would.

Previously, it has `searchString = "#{selector}#{keystrokes}#{command}#{source}"`, and the filter checks if the input is a substring of it.
Now, input is properly trimmed and split, and filter checks if each keyword is in the `searchString`.

jeancroy:
perhaps something like  "allow to search keybinding using multi-keyword query" ?

octref:
@lee-dohm 
Hey, I updated the first comment to include a detailed description of the problem, if you want to review it.
I was mis-using the word "fuzzy matching". This isn't doing fuzzy matching, and fuzzy matching wouldn't work well here.

lee-dohm:
Does this new implementation cover out-of-order-but-still-hyphen-separated modifier key names? So `alt-cmd` and `cmd-alt` would both match `alt-cmd-c`?

octref:
@lee-dohm 
No. `cmd-alt` won't match but `cmd alt` would.
Should I implement it though?
I feel this thing shouldn't go too complex. If you are breaking `cmd-alt` to two keywords to do the filtering, then `cmd-alt-t` also needs to be broken down, and `t` could match a lot other things.
And I can only imagine people filter with precise keybindings to solve their keybinding conflicts. So better to leave input with `-` unmodified, so they can find exactly what they are after.

lee-dohm:
Thanks for this!

PR #778, opened 2016-04-28
Dynamically update the updates tooltip

### LINKED ISSUE #511: Status bar icon persists after installing updates
Settings View's status bar icon persists after installing updates if you installed updates by opening Settings View from the keybinding, menu item, or command palette, rather than by clicking on the status bar icon. The only way to dismiss the status bar icon is by clicking it afterwards, or reloading the window.

50Wliu:
Instead of removing the updates tooltip as soon as it gets clicked, dynamically update the number of updates left in the tooltip as individual packages get updated.  When there are no outdated packages left, the tooltip will disappear.

Closes #511

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -57,6 +57,7 @@ module.exports =
     packageManager = null
 
   consumeStatusBar: (statusBar) ->
+    packageManager ?= new PackageManager() # TODO: Why is this needed when the package manager is initialized at the top of the file?
     Promise.all([packageManager.getOutdated(), packageManager.getInstalled()]).then (values) =>
       updates = values[0]
       allPackages = values[1]
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -78,6 +78,7 @@ module.exports =
     settingsView = new SettingsView(params)
 
   showDeprecatedNotification: (packages) ->
+    return unless packages.user?
     deprecatedPackages = packages.user.filter ({name, version}) ->
       atom.packages.isDeprecatedPackage(name, version)
     return unless deprecatedPackages.length
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -53,11 +53,11 @@ module.exports =
 
   consumeStatusBar: (statusBar) ->
     Promise.all([packageManager.getOutdated(), packageManager.getInstalled()]).then (values) ->
-      updates = values[0].length
+      updates = values[0]
       allPackages = values[1]
-      if updates > 0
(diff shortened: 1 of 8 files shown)


50Wliu:
This now actually works (yay!).

Something else I'd like to do though is also update the update count when packages are detected to be outdated.  This will require an event like `package-update-available`, but then the view also needs to cache the list of outdated packages so that subsequent calls to `getOutdated()` doesn't keep increasing the update count.  Working on this now, but consider what I have right now reviewable.

lee-dohm:
Looks good to me. I haven't had a chance to test it. Is this something you're going to continue working on? Or is it something you'd like to merge now and then do the other part as a separate PR?

50Wliu:
@lee-dohm I managed to get the other part working, so I guess so!  The update count now shows up whenever updates are available, not only when the Settings View first initializes (because that's just boring).  I'll start working on the specs, which knowing me might take a long time.

Basic testing (setting `CACHE_EXPIRY` to 1 and downgrading multiple packages, checking for updates, updating them, reloading the window, etc) didn't catch anything unusual.

lib/package-manager.coffee
```diff
@@ -123,6 +123,9 @@ class PackageManager
           value: packages
           expiry: Date.now() + @CACHE_EXPIRY
 
+        for pack in packages
+          @emitPackageEvent 'update-available', pack
```
50Wliu: This could also be `outdated` to match the function name.

lee-dohm:
I'm still planning to take a look at this. This week has been bonkers. I've got it on my todo list for Monday.

lee-dohm:
This looks good except for the tests that are still marked `TODO` and have no body :grinning:

50Wliu:
No longer WIP.  Specs have been finished.

lib/main.coffee
```diff
@@ -51,32 +54,32 @@ module.exports =
...
     settingsView?.remove()
     settingsView = null
+    packageManager = null
 
   consumeStatusBar: (statusBar) ->
-    PackageManager = require './package-manager'
-    packageManager = new PackageManager()
-    Promise.all([packageManager.getOutdated(), packageManager.getInstalled()]).then (values) ->
-      outdatedPackages = values[0]
+    packageManager ?= new PackageManager() # TODO: Why is this needed when the package manager is initialized at the top of the file?
```
damieng: Does it error without this?

50Wliu: All the relevant specs (except the first one) fail because packageManager is null without this line.

PR #786, opened 2016-05-15
Avoid duplicating identical setting values

Alhadis:
I have my `config.cson` [tracked by version control](https://github.com/Alhadis/Atom-PhoenixTheme/blob/master/.config/config.cson), and whenever I open the settings panel of a grammar package, I get junk like this being added:

```coffee
".awk.source":
  editor:
    tabLength: 4
".fixed.fortran.source":
  editor:
    tabLength: 4
".fortran.free.source":
  editor:
    tabLength: 4
".fortran.modern.source":
  editor:
    tabLength: 4
".fortran.preprocessor.source":
  editor:
    tabLength: 4
".fortran.punchcard.source":
  editor:
    tabLength: 4
```

This is obviously redundant, because my preferred `tabLength` setting (set globally) is already 4. Unsurprisingly, if I set it to Atom's claustrophobia-inducing default of 2, the setting pollution goes away.

This commit fixes that.

These "clones" can be ascertained programmatically by running `atom.config.getRawScopedValue([".scope.whatever"], "editor.tabLength")`, which will return a value other than `undefined` if it's been replicated in the user's config.

Oddly, this doesn't work in the test suite, which *always* returns `undefined`, even when a replicated setting should've been spawned. Ergo, I haven't been able to write a spec that faithfully tests this.

### DIFF
--- a/lib/settings-panel.coffee
+++ b/lib/settings-panel.coffee
@@ -120,7 +120,7 @@ class SettingsPanel extends CollapsibleSectionPanel
 
   set: (name, value) ->
     if @options.scopeName
-      if value is undefined
+      if value is undefined or `value == atom.config.get(name)`
         atom.config.unset(name, scopeSelector: @options.scopeName)
       else
         atom.config.set(name, value, scopeSelector: @options.scopeName)


Alhadis:
@lee-dohm Any reason why backticks are forbidden? Since CoffeeScript [doesn't support/allow](http://stackoverflow.com/questions/7032398/does-coffeescript-allow-javascript-style-equality-semantics) non-strict comparison, there's no way to test equality between two values without manually checking each possible type. Which in this case is necessary, as one value is supplied to the method as a string, and the value returned by `atom.config.get` is a number.

lib/settings-panel.coffee
```diff
@@ -120,7 +120,7 @@ class SettingsPanel extends CollapsibleSectionPanel
 
   set: (name, value) ->
     if @options.scopeName
-      if value is undefined
+      if value is undefined or `value == atom.config.get(name)`
```
thomasjo: First of all, we don't allow the use of "embedded JavaScript", so you'll have to drop the backticks here regardless of justification. Which brings me to the second part; why are you using backticks here? Are you actively trying to avoid `===`?

Alhadis: @thomasjo I literally added a comment just after you commented, which explains why.

lee-dohm:
I'm not sure what the rationale is exactly on no backticks. I'll have to chat with the team and get back to you.

Alhadis:
@lee-dohm Any update?

Alhadis:
Never mind. I'd prefer the Atom team focus [on](https://github.com/atom/tree-view/pull/825) [far](https://github.com/atom/tree-view/pull/833) [more](https://github.com/atom/tree-view/pull/859) [important](https://github.com/atom/find-and-replace/pull/726) PRs instead.

PR #794, opened 2016-05-21
Minor fix for before pseudo element

Connormiha:
CSS code style. `::` is more preference, than `:`

```css
/* CSS3 syntax */
element::before { style properties }

/* CSS2 obsolete syntax (only needed to support IE8) */
element:before  { style properties }
```

https://developer.mozilla.org/en/docs/Web/CSS/::before


50Wliu:
Thanks!

PR #799, opened 2016-06-21
Added styles to status bar notification to make the interaction more obvious

justo:
All other interactable items in the status bar have a pointer cursor and text underline on hover. It makes sense to apply that style to the package update notification as well since clicking it opens a new tab.

This is similar to https://github.com/atom/about/pull/32, which adds a pointer to the squirrel icon.

I understand there is a PR to _remove_ all `cursor:pointers` but until that happens I think it's worth having consistency across the status bar modules. [The branch](https://github.com/atom/atom/pull/7698) has definitely gotten a little stale.

### DIFF
--- a/styles/settings-view.less
+++ b/styles/settings-view.less
@@ -698,6 +698,13 @@
   }
 }
 
+.package-updates-status-view {
+  cursor: pointer;
+  &:hover .available-updates-status {
+    text-decoration: underline;
+  }
+}
+
 .available-updates-status {
   padding-left: 2px;
 }
--- a/styles/settings-view.less
+++ b/styles/settings-view.less
@@ -698,10 +698,11 @@
   }
 }
 
-.package-updates-status-view {
-  cursor: pointer;
-  &:hover .available-updates-status {
-    text-decoration: underline;
+.package-updates-status-view
+  &:hover {
+    .available-updates-status {
+      text-decoration: underline;
+    }
   }
 }
 
--- a/styles/settings-view.less
+++ b/styles/settings-view.less
@@ -698,12 +698,8 @@
   }
 }
 
-.package-updates-status-view
-  &:hover {
-    .available-updates-status {
-      text-decoration: underline;
-    }
-  }
+.package-updates-status-view:hover {
+  text-decoration: underline;
 }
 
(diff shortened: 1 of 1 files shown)


justo:
Not sure why the CI build failed ¯\_(ツ)_/¯

lee-dohm:
I agree with the underline. I disagree with the `cursor: pointer` though. The direction we want to go is to not have `cursor: pointer`. Going backwards is not something we want to do. If you want consistency, you could submit smaller PRs to fix other status bar widgets to not have `cursor: pointer` :grinning:

justo:
Fair enough. If you think it's worth merging I can remove the `cursor` and leave the underline. I wouldn't mind removing the pointer styles from the other widgets but I don't know if you _really_ want to deal with a ton of PRs for such a minor change.

lee-dohm:
@justo I think it is worth merging, yes.

And yes, lots of small PRs where the goal is clear and it doesn't require a lot of testing is fine :grinning:

@simurai Do you still agree that we should be removing `cursor: pointer`?

simurai:
> Do you still agree that we should be removing `cursor: pointer`?

Yeah, I would say as long as there is at least some other hover indicator for items that are clickable. Most have an underline, but color change could be ok too.

styles/settings-view.less
```diff
@@ -698,6 +698,14 @@
   }
 }
 
+.package-updates-status-view
```
simurai: There is an opening `{` missing at the end.

We could also simplify the selector to:

```css
.package-updates-status-view:hover {
  text-decoration: underline;
}
```

The icon wouldn't get the underline.

justo: Whoops, no idea how I missed that `{`.

You're right about the underline, I'll updated that.

simurai:
This would be ready to 🚢 .

I think the build error is unrelated and happens on master too: https://github.com/atom/settings-view/blob/master/README.md. We can wait a little and see if it gets fixed.

justo:
Alright, glad it wasn't me. I didn't think that css would break the build haha.

simurai:
And ✅  👉  🚢

PR #800, opened 2016-06-23
[WIP] Packages star features

bastilian:
This PR is to implement #141 and add features to star packages and make them available to install.
- [x] Clear results on empty search
- [x] Make starred packages available in "Install" panel
- [x] Show package stars in PackageCard
- [x] Implement token authentication for `apm` (see [auth.coffee](https://github.com/atom/apm/blob/master/src/auth.coffee))
- [x] Allow to star/unstar packages in PackageCard
- [x] Limit starred packages to a reasonable number and allow showing more on demand.
- [x] Reload starred packages on star and unstar
- [x] Show hint when no packages are starred and hint to star some
- [x] Polish style
- [x] Ensure responsive interaction
- [ ] Specs for starred packages feature

As a start the first commit adds a starred packages section to "Packages"

screen shot 2016-06-23 at 13 05 13

### DIFF
--- a/lib/atom-io-client.coffee
+++ b/lib/atom-io-client.coffee
@@ -5,6 +5,19 @@ glob = null    # defer until used
 request = null # defer until used
 DefaultRequestHeaders = {'User-Agent': navigator.userAgent}
 
+try
+  keytar = require 'keytar'
+catch error
+  # Gracefully handle keytar failing to load due to missing library on Linux
+  if process.platform is 'linux'
+    keytar =
+      findPassword: ->
+      replacePassword: ->
+  else
+    throw error
+
+tokenName = 'Atom.io API Token'
+
 module.exports =
 class AtomIoClient
   constructor: (@packageManager, @baseURL) ->
@@ -22,6 +35,29 @@ class AtomIoClient
         callback null, cached
       else
         @fetchAndCacheAvatar(login, callback)
+
+  # Get the Atom.io API token from the keychain.
+  #
+  # callback - A function to call with string token as an argument.
+  getToken: (callback) ->
+    if token = keytar.findPassword(tokenName, 'atom.io')
+      callback(token) if callback
+      return
+
+    if token = process.env.ATOM_ACCESS_TOKEN
+      callback(token) if callback
+      return
+
+    if callback
+      callback """
+        No Atom.io API token found.
+      """
+
+  # Save the given token to the keychain.
+  #
+  # token - A string token to save.
+  saveToken: (token) ->
+    keytar.replacePassword(tokenName, 'atom.io', token)
 
   # Public: get a package from the atom.io API, with the appropriate level of
   # caching.
--- a/lib/atom-io-client.coffee
+++ b/lib/atom-io-client.coffee
@@ -49,9 +49,7 @@ class AtomIoClient
       return
 
     if callback
-      callback """
-        No Atom.io API token found.
-      """
+      callback()
 
   # Save the given token to the keychain.
   #
--- a/lib/install-panel.coffee
+++ b/lib/install-panel.coffee
@@ -45,7 +45,7 @@ class InstallPanel extends ScrollView
           @div outlet: 'loadingMessage', class: 'alert alert-info icon icon-hourglass'
           @div outlet: 'featuredContainer', class: 'container package-container'
 
-      @div class: 'section packages', =>
+      @div outlet: 'starredPackagesSection', class: 'section packages', =>
         @div class: 'section-container', =>
           @div outlet: 'starreedHeading', class: 'section-heading icon icon-star', 'Starred Packages'
           @div outlet: 'starreedErrors'
@@ -68,7 +68,15 @@ class InstallPanel extends ScrollView
     @searchType = 'packages'
     @handleSearchEvents()
 
-    @loadStarredPackages()
+    if @showStarredPackages()
+      @loadStarredPackages()
+    else
+      atom.config.observe 'settings-view.enableStarredPackages', (enabled) =>
+        if enabled
+          @loadStarredPackages()
+
+      @starredPackagesSection.hide()
+
     @loadFeaturedPackages()
 
   dispose: ->
@@ -76,6 +84,9 @@ class InstallPanel extends ScrollView
 
   focus: ->
     @searchEditorView.focus()
+
+  showStarredPackages: ->
+    @starredPackages ?= atom.config.get('settings-view.enableStarredPackages')
 
   handleSearchEvents: ->
     @disposables.add @packageManager.on 'package-install-failed', ({pack, error}) =>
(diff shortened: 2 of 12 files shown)


simurai:
Seeing that screenshot got me thinking: Currently starring a package on atom.io is more considered as a "like", well IMO. Having them inside Atom's settings might change that to a "bookmarking" (so you can find/install it later). Kinda the opposite of the fav ⭐  to like ❤️  change on Twitter.

Also, there are probably people that have starred tons of packages, for those it will fill up that page quite a bit. And some people might start to un-star packages to save space. So...

- Should it be limited to only show your 20 most recently starred packages? With a link to https://atom.io/users/name/stars for the rest?
- If you star a package, does the scroll position jump because a new package card gets added above? Depends how quickly this happens.
- Should we show starred packages in the "Install" panel instead? So that the "Packages" panel is reserved for only installed packages.

bastilian:
True, that could change the meaning to star a project.

Limiting it to the most recent packages is a good idea as well as moving them to the install panel. That should eliminate the second issue as well.

However I would allow to load more starred packages in Atom, rather than linking to the website, that way we can serve those as well that want to use it as a tool to have all their packages ready for install.

bastilian:
I'm gonna implement clearing the search results when an empty string is the query, to make it possible to return to a clean screen again with featured and starred packages visible. This will also address #590.

I've also opened an issue for atom to introduce a clear function for the TextEditor to clear results with a click [#atom/atom#12048](/atom/atom/issues/12048)

bastilian:
While adding the starred list to the install panel I wanted it to behave like the featured packages and make it switch between themes and packages, but it is not yet possible to get only packages from `apm stars`.

It would be possible to implement the filter here, but I think it should rather be added to `apm`.
I opened an [issue](/atom/apm/issues/585) there to add the option, will implement it and if/when it is merged we can have the starred list behave like the featured one.

bastilian:
It's working

It's still WIP, but it works and you can star and unstar packages in Atom. 

To not alarm users with an unexpected "Keychain access" prompt, I added a settings-view setting:
screen shot 2016-06-29 at 13 29 51

Once it is enabled it will ask for permission to access the keychain and will show the "Starred Packages" section in "Install". In case there is not yet a access token set (via `apm login`) it will ask for one:
screen shot 2016-06-29 at 13 24 39

If the key is set you'll see you packages right in Atom. Woohoo! 🎉  
screen shot 2016-06-29 at 13 23 35

It also allows to star and unstar packages under "Featured Packages" and search results.
screen shot 2016-06-29 at 13 23 23

As said, it works, but there are still some issues that need to be addressed. Clicking on the star triggers under certain condition weird jumping behaviour, which I need to figure.

Another thing is entering the token. To be consistent with the other settings, setting the token will trigger loading the starred packages right away, this feels a bit weird.
A better solution might be to get #458 ready for merge and show the setting in the settings-view settings if starred packages are enabled.

I'll add specs next (maybe i can also fix that random failing spect), but feel free to test it how it feels and let me know what you think!

Ben3eeE:
Typo in first image? "It requires access to your systems ***keychachain***"
I like this and will give it a round of testing this week!

bastilian:
I probably listened to bowies "changes"... hahaha. fixed it. :)

bastilian:
I've added a spec to verify that the starred packages are shown when it is enabled, but I can't figure out why it fails, even though the enabled setting is true and the section is not getting hidden in the process.

Anyone able to take a look and help me figure it out?

The spec in question: https://github.com/atom/settings-view/pull/800/files#diff-93e1268d244602e563a3af1ce54c6f0eR111

bastilian:
If there are now more then 10 packages it'll show a "Show X more" link instead of loading them all into the view:

screen shot 2016-07-08 at 19 52 18

I've also now implemented the reloading of the starred packages when one gets starred or unstarred and added a proper loading message and a hint that if no packages are starred yet.

screen shot 2016-07-08 at 19 45 09

screen shot 2016-07-08 at 19 42 32

I'm happy with how the "Show more" looks like, but still need to style the "No package" hint.

There is an issue still when star/unstar a package and refresh the list, that the packages are not updated yet. I need to investigate where the culprit lies and find a way to ensure the list in the interface is updated correctly.

This and writing the specs and the PR should be ready for extensive testing.

Ben3eeE:
> I like this and will give it a round of testing this week!

Been busy and didn't have time until now, sorry about that. I did some basic testing and found some things and have some comments.

- Show more is shown before you enter account token.
show more shown early
- Starred packages list does not update at all for me until I restart Atom. But I think you already knew this based on your previous comment.
- Featured packages list does not star packages correctly, see image, `github-atom-light-syntax` is not starred in featured packages but it is starred and shown correctly in starred packages.
featured unstarred starred starred
- No way to star/unstar packages from packages tab. I looked for this because my first instinct was to star some of my installed packages for easier find if I install Atom on another computer, which is one great use for this. Now I had to check packages tab and go and search for each package individually, which is tedious.
- The list in the install tab is very long vertically and annoying to scroll if you want to go to the starred section. Perhaps each of these 3 sections (Search, Featured, Starred) should collapse in the same way as the different sections do in the packages tab?
- Both featured packages and starred packages have the same star icon next to the text, sometimes when scrolling through to find starred packages I found the star and thought I was at the starred packages section when I was not.

starred packages icon
featured packages icon

bastilian:
@Ben3eeE: thanks for giving it a test run! I'll look into the issue you found and fix them up one after another. :)

Ben3eeE:
@bastilian Let me know when you want more testing.

bastilian:
@Ben3eeE cool. I (finally) had time now to iron out a few things.

It should now star and unstar packages correctly when it's enabled and with a token.
The list is now also updating correctly and there should not be any need to restart atom to update it.

I did not yet enable it in the "Packages" section for now, as "Install" poses a good testing ground and fulfils the primary purpose of this, which is to add a way to allow installing of starred packages.

Where and how this feature is added is still up for discussion.

I added basic specs to test the feature, but still needs some to verify functionality.
But, you (and everyone who wants) is free to test it. 

Thank you!

Ben3eeE:
> Where and how this feature is added is still up for discussion.

Applies to the discussions in https://github.com/atom/settings-view/issues/825 maybe should ask there for opinions.

bastilian:
Once #825 has come to a conclusion, it is just a matter of flipping the switching and turning it on.
But as said for testing now the Install tab should work just fine.

I would also love @simurai's feedback on this. :)

simurai:
> Once #825 has come to a conclusion

Maybe starred packages could have their own tab? Then it doesn't interfere with whatever comes out of #825.

screen shot 2016-10-19 at 1 56 44 pm

damieng:
It seems discussion on this stopped and now it's out of date. Is there any interest in getting the discussion going again?

bastilian:
I'm closing this.

PR #802, opened 2016-06-24
Fix overflowing of the keybindings panel

simurai:
Same issue as https://github.com/atom/atom/pull/11866

#### Before

Content that overflows the edge gets cut off and you can't scroll.

screen shot 2016-06-24 at 3 33 02 pm

#### After

Content is "responsive" and the table is scrollable again.

screen shot 2016-06-24 at 3 32 19 pm

### DIFF
--- a/styles/settings-view.less
+++ b/styles/settings-view.less
@@ -458,6 +458,7 @@
     flex-flow: column;
     display: flex;
     position: relative;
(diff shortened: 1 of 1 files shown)


simurai:
Because this is somewhat urgent, gonna merge this even though "AppVeyor build failed".

PR #808, opened 2016-06-30
Move the editor settings to their own tab

damieng:
The Settings tab within Settings was getting very long and is already broken into Core and Editor.

This PR renames the current tab Core and moves Editor out to its own tab, as shown;;

image


nathansobo:
I'm in favor. Seems like it will be more usable. Thanks for adding tests for the new panel.

nathansobo:
Thanks for taking the initiative here ✨

Issue #744, opened 2016-03-03
Command to Show Core Settings


JamesCoyle:
There doesn't seem to be a command which will switch to the core settings tab while the settings view is open. Would be kinda nice to have as it is the only one which is missing.

damieng:
Is included in #811

PR #811, opened 2016-07-06
Add new System tab for Windows shell integration options

damieng:
Adds a new System tab on Windows that allows configuration of the OS shell integration options for things like the file handler, file context menu and folder context menu.

### DIFF
--- a/lib/system-windows-panel.coffee
+++ b/lib/system-windows-panel.coffee
@@ -19,7 +19,7 @@ class SystemPanel extends ScrollView
                       @input outlet: 'fileHandlerCheckbox', id: 'system.windows.file-handler', type: 'checkbox'
                       @div class: 'setting-title', 'Register as file handler'
                       @div class: 'setting-description', =>
-                        @raw("Show #{WindowsShell.appName} in the \"Open with\" application list for easy association with file types.")
+                        @raw("Show #{WinShell.appName} in the \"Open with\" application list for easy association with file types.")
               @div class: 'control-group', =>
                 @div class: 'controls', =>
                   @div class: 'checkbox', =>
@@ -27,7 +27,7 @@ class SystemPanel extends ScrollView
                       @input outlet: 'fileContextMenuCheckbox', id: 'system.windows.shell-menu-files', type: 'checkbox'
                       @div class: 'setting-title', 'Show in file context menus'
                       @div class: 'setting-description', =>
-                        @raw("Add \"Open with #{WindowsShell.appName}\" to the File Explorer context menu for files.")
+                        @raw("Add \"Open with #{WinShell.appName}\" to the File Explorer context menu for files.")
               @div class: 'control-group', =>
                 @div class: 'controls', =>
                   @div class: 'checkbox', =>
@@ -35,7 +35,7 @@ class SystemPanel extends ScrollView
                       @input outlet: 'folderContextMenuCheckbox', id: 'system.windows.shell-menu-folders', type: 'checkbox'
                       @div class: 'setting-title', 'Show in folder context menus'
                       @div class: 'setting-description', =>
-                        @raw("Add \"Open with #{WindowsShell.appName}\" to the File Explorer context menu for folders.")
+                        @raw("Add \"Open with #{WinShell.appName}\" to the File Explorer context menu for folders.")
 
   initialize: ->
     super
@@ -43,11 +43,11 @@ class SystemPanel extends ScrollView
     WinShell.fileContextMenu.isRegistered (i) => @fileContextMenuCheckbox.prop('checked', i)
     WinShell.folderContextMenu.isRegistered (i) => @folderContextMenuCheckbox.prop('checked', i)
 
-    @fileHandlerCheckbox.on 'click', (e) -> setRegistration WinShell.fileHandler, e.target.clicked
-    @fileContextMenuCheckbox.on 'click', (e) -> setRegistration WinShell.fileContextMenu, e.target.clicked
-    @folderContextMenuCheckbox.on 'click', (e) ->
-      setRegistration WinShell.folderContextMenu, e.target.clicked
-      setRegistration WinShell.folderBackgroundContextMenu, e.target.clicked
+    @fileHandlerCheckbox.on 'click', (e) => @setRegistration WinShell.fileHandler, e.target.checked
+    @fileContextMenuCheckbox.on 'click', (e) => @setRegistration WinShell.fileContextMenu, e.target.checked
+    @folderContextMenuCheckbox.on 'click', (e) =>
+      @setRegistration WinShell.folderContextMenu, e.target.checked
+      @setRegistration WinShell.folderBackgroundContextMenu, e.target.checked
 
   setRegistration: (option, shouldBeRegistered) ->
     if shouldBeRegistered
--- a/lib/system-windows-panel.coffee
+++ b/lib/system-windows-panel.coffee
@@ -1,0 +1,102 @@
+{ScrollView} = require 'atom-space-pen-views'
+SettingsPanel = require './settings-panel'
+Registry = require 'winreg'
+Path = require 'path'
+
+exeName = Path.basename(process.execPath)
+appPath = "\"#{process.execPath}\""
+appName = exeName.replace('atom', 'Atom').replace('beta', 'Beta').replace('.exe', '')
+
+fileHandlerRegistry = {
+  key: "\\Software\\Classes\\Applications\\#{exeName}",
+  parts: [{key: 'shell\\open\\command', name: '', value: "#{appPath} \"%1\""}]
+}
+
+contextRegistryParts = [
+    {key: 'command', name: '', value: "#{appPath} \"%1\""},
+    {name: '', value: "Open with #{appName}"},
+    {name: 'Icon', value: "#{appPath}"}
+]
+
+contextFileRegistry = {
+  key: "\\Software\\Classes\\*\\shell\\#{appName}",
+  parts: contextRegistryParts
+}
+
+contextFolderRegistry = {
+  key: "\\Software\\Classes\\Directory\\shell\\#{appName}",
+  parts: contextRegistryParts
+}
+
+contextFolderBackgroundRegistry = {
+  key: "\\Software\\Classes\\Directory\\background\\shell\\#{appName}",
+  parts: JSON.parse(JSON.stringify(contextRegistryParts).replace('%1', '%V'))
+}
+
+module.exports =
+class SystemPanel extends ScrollView
+  @content: ->
+    @div tabindex: 0, class: 'panels-item', =>
+      @form class: 'general-panel section', =>
+        @div class: 'settings-panel', =>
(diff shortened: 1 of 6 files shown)


lib/system-windows-panel.coffee
```diff
@@ -0,0 +1,102 @@
...
+  parts: [{key: 'shell\\open\\command', name: '', value: "#{appPath} \"%1\""}]
+}
+
+contextRegistryParts = [
+    {key: 'command', name: '', value: "#{appPath} \"%1\""},
+    {name: '', value: "Open with #{appName}"},
+    {name: 'Icon', value: "#{appPath}"}
+]
+
+contextFileRegistry = {
```
nathansobo: Seems like these should be SHOUTING_CASE style constants.

lib/system-windows-panel.coffee
```diff
@@ -0,0 +1,102 @@
...
+  key: "\\Software\\Classes\\*\\shell\\#{appName}",
+  parts: contextRegistryParts
+}
+
+contextFolderRegistry = {
+  key: "\\Software\\Classes\\Directory\\shell\\#{appName}",
+  parts: contextRegistryParts
+}
+
+contextFolderBackgroundRegistry = {
```
nathansobo: It's weird that these are called "registry"... Isn't it more like "registry entry" or something? Isn't the entire thing you're accessing called "The Registry".

nathansobo:
Aside from some minor stuff this all looks good.

lib/system-windows-panel.coffee
damieng: In other languages that's been traditionally reserved for compile-time constants - do we or Coffeescript do something different?

lib/system-windows-panel.coffee
damieng: contextFolderRegistrySettings perhaps?

nathansobo: Yeah, kinda long... What about "settings".

nathansobo: I mean dropping the word "registry"

damieng: I chose registry to differentiate from regular config-file settings that settings-view normally deals with.

lib/system-windows-panel.coffee
nathansobo: Since there is no compile time we do... In a lot of places we use CamelCase but that's actually not standard. In JS these kinds of constants declared at the top of the file are typically ALL_CAPS

damieng:
A couple of questions;

1. Do we want to add a command for this? settings-view:system perhaps or settings-view:system-windows ?
2. How do we handle moving the registry stuff into Atom? Do we set a min-version on this version of settings-view or just make it only display the system tab if you have Atom >= 1.10 ?

damieng:
This will fix both atom/atom#5901 and atom/atom#9247

nathansobo:
@damieng We can use version detection or just look for the presence of the desired API. What is the core API you're imagining Atom exposing?

nathansobo:
As for the command, I don't feel strongly either way. Should probably make it `system` rather than `system-windows` in case we add the tab for other platforms. Should probably only add the tab and the command to the registry if we're on a platform that actually has a system tab. Are there commands for other settings tabs? If so, I'd say blend in. If not, I'd say don't do it.

lee-dohm:
Applies to #744.

nathansobo:
🌸🌸🌸

jerone:
I just updated to the latest beta (Atom 1.10.0-beta0, Settings-view 0.241.0) on Windows and noticed that the default value for this new pane don't lineup with the situation before updating.

Here's a screenshot of the new pane, with all checkboxes unchecked, while everything should be checked (on the right proof that the second checkbox should be checked):
untitled

<sub>ps. If required, I can make this a separate issue...</sub>

damieng:
That is odd. Did you have a beta version before?

If you try toggling the option does it disappear?

jerone:
[**@damieng**](https://github.com/damieng) commented on [1 aug. 2016 21:33 CEST](https://github.com/atom/settings-view/pull/811#issuecomment-236682978 "2016-08-01T19:33:40Z - Replied by Github Reply Comments"):
> That is odd. Did you have a beta version before?

Yes, running Beta's from the beginning.

> If you try toggling the option does it disappear?

No, it appears the options (at least "Show in file context menus") doesn't seem to do anything.

First I made sure those settings weren't in my `config.cson`; which they weren't.
Second, I tried a clean profile, which made no difference.

damieng:
Can you take a screenshot of the following registry key branch;

`HKEY_CURRENT_USER\SOFTWARE\Classes\directory\shell`

In there I would expect to see a folder 'Atom Beta' with a sub-branch called `command` that has a default value that points to your atom beta exe.

Also are you on a corporate network that might have restrictions on registry access via group policies?

jerone:
[**@damieng**](https://github.com/damieng) commented on [1 aug. 2016 22:11 CEST](https://github.com/atom/settings-view/pull/811#issuecomment-236692843 "2016-08-01T20:11:01Z - Replied by Github Reply Comments"):
> Can you take a screenshot of the following registry key branch;
> 
> `HKEY_CURRENT_USER\SOFTWARE\Classes\directory\shell`
> 
> In there I would expect to see a folder 'Atom Beta' with a sub-branch called `command` that has a default value that points to your atom beta exe.

2016-08-01 22_19_47-registry editor

`Command` folder only has default, which indeed points to current beta.
`Atom` parent folder has default and `Icon` value.

> Also are you on a corporate network that might have restrictions on registry access via group policies?

No corporate network. Have full admin rights.

jerone:
@damieng 

Since the last communication I restarted my laptop and updated to the last Atom beta (1.10.0-beta1) and now I have two Atom icons in my context-menu.

This has been confirmed by the Registry:
2016-08-05 20_46_09-add new system tab for windows shell integration options by damieng pull reque

The first Atom entry has an "Icon" item which refers to `"C:\Users\jerone\AppData\Local\atom\app-1.10.0-beta0\atom.exe"`.
The second Atom Beta also has an "Icon" item which refers to `"C:\Users\jerone\AppData\Local\atom\app-1.10.0-beta1\atom.exe"` (notice the `beta1` difference). 
Same goes for the "command" sub-entry.

damieng:
Atom Beta today "upgrades" Atom and then it gets in this orphaned state. 

If you want to install Atom Beta side-by-side you'll need to grab the "atom-windows.zip" and manually unpack it somewhere.

Alternatively if you want to use the beta without the core Atom I'd recommend uninstalling the main Atom installation first.

Ideally the Atom Beta installer is going to allow you to install beta and non-beta side by side but there is a little work to do there yet.

jerone:
I'm **not** running multiple Atom versions. Always run the latest beta since the beginning they were available.

damieng:
I never said you were running multiple versions.

What version did you have on your machine before you upgraded?

jerone:
The previous beta version: Atom 1.10.0-beta0

damieng:
This was done to allow Atom and Atom Beta to be able to run side-by-side in the future but yes the upgrade experience of the current beta installer isn't smart enough to clean up pre-1.10-beta1 states.  

You could just delete 'Atom' keys from Directory, * and Directory/Background for now to get you out of the state. Going forward it shouldn't happen again unless you go from Atom to Atom Beta before the installers allow side-by-side installation.

jerone:
Oke, so it's not really a bug, more an undesired behavior or missing implementation.

I can deal with double Atom entries in my context-menus, but does this create issues with duplicate values for the file handler option?

Ben3eeE:
@damieng The values in the settings view systems tab got cleared for me when I updated today. The context menu items still show up and work perfectly fine but all of the settings in the systems tab are unchecked.
I have only had the stable release installed on this computer.
Yesterday in Atom 1.10.1 I checked two of the options (Both context menu ones) and now when I open the settings view after updating to 1.10.2 all of the values are unchecked.

I will keep an eye open to see if they change again next update.

jerone:
Same issue for me again. [Previous time](https://github.com/atom/settings-view/pull/811#issuecomment-236678606) it was at home where I have the beta version running, so I could understand if somethings breaks there. 
However, yesterday I updated at work where I have the release version running (1.10.1). All three options got cleared after the update. Enabling them made all options work again.
But then today, I updated at work to 1.10.2 and the option "_Show in file context menus_" & "_Show in folder context menus_" got cleared. While they context-menu option still works in Explorer. I'm 100% confident I checked them in 1.10.1.

atom system settings

_Home: Windows 7 x64_
_Work: Windows 8.1 x64_

damieng:
There is a known issue that beta versions will interfere with non-beta when it comes to registering in the shell.  This is because both executables are named atom.exe which is used in the registry key to register the application.

If you didn't use the beta or dev builds then it's a different problem we should investigate.  The Squirrel installer should be automatically upgrading any shell integration you have between versions.

(As an aside the settings themselves are not actually cleared - these checkboxes are dynamically evaluated by comparing the current atom executable against whether that executable is set to do this in the Windows registry.  This means they appear unchecked if you move the executable such as renaming your profile.)

Ben3eeE:
The shell integration worked perfectly fine after updating with the squirrel updater from 1.10.1 to 1.10.2 but the checkbox state in the settings-view is not evaluated correctly after the update, it is unchecked after the update when it should be checked. I did not move the executable.

I did have problems where the shell integration did not work when going from 1.9.9 to 1.10 which was fixed by changing these settings in the settings view which is how I know I had them set before updating to 1.10.2.
There was another report of someone having the same problems as I had on discuss: https://discuss.atom.io/t/cant-open-folder-using-atom-by-right-click-open-with-atom/33038/6
and also in the referenced issue https://github.com/atom/atom/issues/11298 there seems to be similar problems.

damieng:
The squirrel updater script includes steps to adjust the paths in the registry to the new version.  If it ran correctly the checkboxes should still be checked.

Can you edit the file apm.cmd in the %localappdata%\atom\bin folder and see what version number it contains?

Ben3eeE:
```
@echo off
"%~dp0\..\app-1.10.2\resources\app\apm\bin\apm.cmd" %*
```

damieng:
Yeah the script either isn't working or isn't being run.  There are also reports in prior versions of the bin\apm.cmd not being updated (it is in yours).  Wondering if the root cause is the same or just both happen to be in the squirrel updater area.  Unfortunately this part of the update process has no facility to log to squirrel setup so might need some new logging if I can't repro it later today.

OliverRC:
@damieng so I am assuming there is some script/command/process that happens post installation to update all necessary registry entries, etc to the latest folder. Is this something I can invoke myself to "fix" an update that didn't seem to work smoothly.

I'd been keen to see what exactly is causing it to fail, but something must be preventing the update (Squirrel?) from correctly updating the registry.

damieng:
Some work was done late in the cycle to try and enable Atom and Atom Beta to be run side-by-side but broke the upgrade process https://github.com/atom/atom/commit/9b233963d3836af25fc0b798e964d16ca5639076  

It's now fixed in master and should go out in any other future updates so shouldn't see this problem again.

PR #816, opened 2016-07-23
Order packages and themes case-insensitively

Alhadis:
Minor detail. Packages that begin with a capital letter appear at the top of the packages and themes list, which looks pretty jarring (more so when everything's already formatted in title-case).

I'm aware NPM (and by convention, APM) force users to use lowercase package names, but this won't stop a developer from linking an unpublished package into their workspace (even if it was just to [test something briefly](https://github.com/Alhadis/Breakshow)).

### DIFF
--- a/lib/themes-panel.coffee
+++ b/lib/themes-panel.coffee
@@ -229,7 +229,15 @@ class ThemesPanel extends CollapsibleSectionPanel
   populateThemeMenus: ->
     @uiMenu.empty()
     @syntaxMenu.empty()
-    availableThemes = _.sortBy(atom.themes.getLoadedThemes(), 'name')
+    availableThemes = atom.themes.getLoadedThemes().sort (a, b) ->
+      nameA = a.name.toLowerCase()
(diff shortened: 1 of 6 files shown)


thomasjo:
> I'm aware NPM (and by convention, APM) force users to use lowercase package names, but this won't stop a developer from linking an unpublished package into their workspace

This says it all to me. I don't understand why it's worth adding this to fix something that is essentially not a Real World™ problem — I find it hard to justify the added complexity.

I'm 👎 on this one. /cc @atom/feedback

Alhadis:
Oh yeah, there are also a minority of packages published with title-case (before lowercase became forced on users). E.g., [this one](https://github.com/bigfive/atom-sublime-select/blob/master/package.json).

Current number of installs: [91,321](https://atom.io/packages/sublime-style-column-selection).

Alhadis:
Still, if this approach is too complex, maybe consider lowercasing the name of each package that's loaded at startup...?

Because whether or not lowercased names are the popular convention, this discrepancy just looks silly.

PR #819, opened 2016-07-26
Unify icons between tabs and titles

damieng:
- Core Settings title now uses icon-settings like sidebar
- System Settings now title-cased
- Install title now uses icon-plus like sidebar
- Theme title now uses icon-paintcan like sidebar

### DIFF
--- a/lib/editor-panel.coffee
+++ b/lib/editor-panel.coffee
@@ -13,7 +13,7 @@ class EditorPanel extends ScrollView
     @loadingElement.remove()
 
     @subPanels = [
-      new SettingsPanel('editor', note: '''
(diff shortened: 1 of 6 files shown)


50Wliu:
Would the [text-size octicon](https://octicons.github.com/icon/text-size/) maybe look better for the Editor Settings icon?

It's nothing major; I just find the current icon a bit underwhelming with file-icons installed:
settings-view-left-panel

damieng:
Text-size is not in the version of octicons we use and upgrading is a breaking change as some icons have been removed so.

Some other options would be;

code
file-code
list-ordered
three-bars

PR #822, opened 2016-07-29
Fix #821 Settings-view overflows when pane is narrow

### LINKED ISSUE #821: Settings-view overflows when pane is narrow
environment information:

- Atom    : 1.8.0
- Electron: 0.36.12
- Chrome  : 47.0.2526.110
- Node    : 5.1.1
- settings-view    : 0.241.0
- OS: Arch Linux latest

These days I'm writing my transparent styled ui theme. I found settings-view overflows, when pane is narrow.

transparent_overflow_sm

Though in solid backgound themes we cannot see the overflowed content. I found another weird behavoir:  at `Install Packages` package/theme buttons.

solid_overflow_sm

I think `overflow: auto` should be appropriate for it. And here's the fixed preview. Also weird package/theme buttons go away..

fixed_sm

frantic1048:
Sets `overflow: auto` for settings-view to prevent overflow.

see #821

### DIFF
--- a/styles/settings-view.less
+++ b/styles/settings-view.less
@@ -7,6 +7,7 @@
 
 .settings-view {
   display: flex;
+  overflow: auto;
 
   .breadcrumb {
     margin-bottom: 0;


styles/settings-view.less
```diff
@@ -7,6 +7,7 @@
 
 .settings-view {
   display: flex;
+  overflow: auto;
```
simurai: How about instead of making the whole settings-view scrollable, limit it to only `.settings-view .panels`. Then the left navigation stays fixed. Like this:

```diff
.settings-view .panels {
+  overflow: auto;
}

.settings-view .panels .panels-item {
-  overflow: auto;
}
```

frantic1048: @simurai If so, the left navigation will overflow out of whole pane, when pane is narrow,

simurai: Yeah right, if very narrow (smaller than `180px`). I was thinking that if your pane is that narrow, you don't really want to open the settings in it. But be able to scroll everything might be nice too because it gives more space for the panels.

Let's merge this. 👍 @frantic1048 Thanks.

PR #830, opened 2016-08-17
Fix the problem that tabLength of language scope jumps back to global…

taot:
When I set tabLength to 4 in global editor setting, and set it to 2 in language scope setting (such as coffeescript or scala), the one in language setting immediately jumps back to 4. Even when I set it in config file, it still jumps back. It turned out the settings-view does not look at global settings when resetting to default.

So I made this small patch. However I could not make up a valid unit test for this successfully. I tried to reproduce the issue with a jasmine spec, but with no success. I think I am still not familiar enough with Atom yet. So I plan to look at Atom code. And I will be grateful if someone is familiar with Atom and also willing to give me some advice on the unit test.

### DIFF
--- a/lib/settings-panel.coffee
+++ b/lib/settings-panel.coffee
@@ -106,6 +106,13 @@ class SettingsPanel extends CollapsibleSectionPanel
     params.scope = [@options.scopeName] if @options.scopeName?
     @disposables.add atom.config.observe(name, params, callback)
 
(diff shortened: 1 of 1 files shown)


50Wliu:
I believe this was fixed by #841.

taot:
Yeah, I've noticed it. Thanks a lot!

PR #831, opened 2016-08-21
Height incorrectly set to 100%

tensor5:
Fix height of "Choose a Theme" and "Install Packages" panels when running Atom with recent versions of Electron (currently 1.3.3). See https://github.com/tensor5/arch-atom/issues/38.

screenshot from 2016-08-21 10-25-06

screenshot from 2016-08-21 10-25-19

screenshot from 2016-08-21 10-25-23

screenshot from 2016-08-21 10-25-32

### DIFF
--- a/styles/settings-view.less
+++ b/styles/settings-view.less
@@ -542,7 +542,6 @@
   }
 
   .packages {
-    height: 100%;
     display: flex;
     flex-flow: column;
 


damieng:
Tried this with the built-in themes and with a third-party one (Material) and didn't see any problem with it's removal.

tensor5:
Thanks.

PR #842, opened 2016-09-14
Prioritize close matches in install search results

Ingramz:
Current behavior:
search

This pull request moves packages which names contain the search query to the top, below the exact hit. 
search2

I personally think that this will surface more relevant packages in a partial search but it might bury some popular packages, for instance when searching `file-icons` then `file-type-icons` will rank lower, but it will surface more packages that contain the search term. A better long term solution would be to use a scoring system that would view each word individually.

Feedback would be appreciated.

### DIFF
--- a/lib/install-panel.coffee
+++ b/lib/install-panel.coffee
@@ -125,7 +125,7 @@ class InstallPanel extends ScrollView
       null
 
   performSearch: ->
-    if query = @searchEditorView.getText().trim()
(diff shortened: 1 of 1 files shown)


damieng:
Can we get a spec or two for this to ensure it doesn't get broken?

Ingramz:
Thanks for looking into this. I've been busy with other obligations lately and haven't had chance to deal with pull requests.

PR #844, opened 2016-09-14
Added `--production` flag to apm install

KyleKing:
:racehorse:  The `--production` flag makes installing new packages faster and prevents a bug when packages have tons of dev-dependencies

Nylas N1 uses the same install script as Atom and on N1, installing packages wth a large number of dev-dependencies fails to install. There should be no need to install dev-dependencies for a redistributed package, so this should help speed up the install process and avoid errors such as seen in Nylas N1: https://github.com/nylas/N1/issues/2723#issuecomment-240983812

The same change was merged to the master branch of Nylas and released last weekend: https://github.com/nylas/N1/pull/2741

### DIFF
--- a/lib/package-manager.coffee
+++ b/lib/package-manager.coffee
@@ -309,7 +309,7 @@ class PackageManager
     nameWithVersion = if version? then "#{name}@#{version}" else name
 
     @unload(name)
-    args = ['install', nameWithVersion, '--json']
+    args = ['install', nameWithVersion, '--production', '--json']
 
     errorMessage = "Installing \u201C#{nameWithVersion}\u201D failed."
     onError = (error) =>
--- a/spec/package-manager-spec.coffee
+++ b/spec/package-manager-spec.coffee
@@ -82,12 +82,12 @@ describe "package manager", ->
     it "installs the latest version when a package version is not specified", ->
       packageManager.install {name: 'something'}, ->
       expect(packageManager.runCommand).toHaveBeenCalled()
-      expect(runArgs).toEqual ['install', 'something', '--json']
+      expect(runArgs).toEqual ['install', 'something', '--production', '--json']
 
     it "installs the package@version when a version is specified", ->
       packageManager.install {name: 'something', version: '0.2.3'}, ->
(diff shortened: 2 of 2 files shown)


lib/package-manager.coffee
```diff
@@ -309,7 +309,7 @@ class PackageManager
     nameWithVersion = if version? then "#{name}@#{version}" else name
 
     @unload(name)
-    args = ['install', nameWithVersion, '--json']
+    args = ['install', nameWithVersion, '--production', '--json']
```
KyleKing: ^ the only actual change. The rest are changes to the tests to account for the new apm install flag

KyleKing:
What can I do to help? What kind of tests do you want to see?

KyleKing:
Just wanted to follow up. What do you think about the PR?

damieng:
I think the worry is that this may break some packages.

KyleKing:
Oh okay. I think the only issue is when a package incorrectly lists a critical dependency as a dev- dependency, then the owner would only need to fix the package.json and release an update. What would be the best way to mitigate this risk? Is there a way to output a better error message to help when debugging?

BinaryMuse:
@KyleKing Are you certain that devDependencies are installed via apm? I just checked several of my installed packages inside `~/.atom/packages`, and none of the packages I checked that had `devDependencies` had the associated modules installed in their `node_modules` folder. To double check, I also installed a package I found via my copy of Atom (version 1.14.0-dev-60ce394 with apm 1.15.1), [elm-sm](https://atom.io/packages/elm-sm), which contains [electron-prebuilt as a `devDependency`](https://github.com/xorsnn/elm-sm/blob/13a1f939f5b460cbdf786a28301fdf2a1c29e0aa/package.json#L21-L24), and it wasn't installed.

Also, we do actually depend on dev dependencies being installed when installing directly from a Git remote as often the build step must run as part of installation (these are installed via a different code path down in apm to ensure dev dependencies are installed).

KyleKing:
@BinaryMuse I see the same thing - none of the devdependencies are installed for several of my packages as well. I'll go ahead and close the PR

PR #846, opened 2016-09-16
Remove unused variables

Ingramz:


lib/atom-io-client.coffee
```diff
@@ -181,7 +181,7 @@ class AtomIoClient
 
       for key, children of files
         children.sort()
-        keep = children.pop()
+        children.pop() # keep
```
50Wliu: ?

lib/install-panel.coffee
```diff
@@ -175,7 +171,7 @@ class InstallPanel extends ScrollView
         @searchMessage.hide()
         @searchErrors.append(new ErrorView(@packageManager, error))
 
-  showNoResultMessage: ->
+  showNoResultMessage: (query) ->
```
50Wliu: Nice!

lib/main.coffee
```diff
@@ -60,7 +60,7 @@ module.exports =
       allPackages = values[1]
       if outdatedPackages.length > 0
         PackageUpdatesStatusView = require './package-updates-status-view'
-        packageUpdatesStatusView = new PackageUpdatesStatusView(statusBar, outdatedPackages)
+        new PackageUpdatesStatusView(statusBar, outdatedPackages)
```
50Wliu: Is this and `PackageDetailView` (further below) still needed?

Ingramz: Space pen views call `@initialize` of the extended class from constructor, which seems to be responsible for the magic of presenting itself. https://github.com/atom/settings-view/blob/c06ae51ac092d2a608d4cdff4677eeb2453e3e3d/lib/package-updates-status-view.coffee#L11-L14

lib/atom-io-client.coffee
50Wliu: Also, this can be `children.sort().pop()` now right?

PR #858, opened 2016-09-23
Update README.md and LICENSE.md

torn4dom4n:


50Wliu:
Thanks!

damieng:
Copyrights are supposed to retain the oldest original date btw.

50Wliu:
> Copyrights are supposed to retain the oldest original date btw.

I wasn't sure about that, so I checked atom/atom's LICENSE, which had a similar PR accepted to change the year to 2016.

damieng:
We should probably correct both of them.  The US copyright.gov site says "The year of first publication. " http://www.copyright.gov/circs/circ03.pdf

50Wliu:
Done.  Thanks for the clarification!

torn4dom4n:
I don't know about that. Sorry!

PR #867, opened 2016-10-12
Fix scrolling performance issues in settings-view

easyaspi314:
Part of fix for atom/atom#12949.

Forces Chromium to use the compositor to render scroll containers. This greatly decreases CPU usage when scrolling.

### DIFF
--- a/styles/settings-view.less
+++ b/styles/settings-view.less
@@ -8,6 +8,7 @@
 .settings-view {
   display: flex;
   overflow: auto;
+  will-change: transform;
 
   .breadcrumb {
     margin-bottom: 0;
@@ -372,6 +373,7 @@
     background-color: lighten(@base-background-color, 2%);
     border-right: 1px solid @base-border-color;
     overflow-x: auto;
+    will-change: transform;
 
     .icon:before {
       text-align: center;
@@ -380,6 +382,7 @@
     .panels-packages{
       flex-grow: 1;
       overflow: auto;
+      will-change: transform;
       height: 0;
 
       li + li {
@@ -465,6 +468,7 @@
       flex: 1;
       min-width: 372px; // magic number: fit the Settings, Uninstall and Disable button
       overflow: auto;
+      will-change: transform;
     }
   }
 


simurai:
@easyaspi314 Thanks for looking into this. ❤️ 

Moving the package list into its own layer with `will-change: transform;` does improve scrolling. But I'm a bit concerned about adding it permanently. I think `will-change` is only meant to be used temporarily before something gets animated and should be removed later. I know, in our case pretty hard to predict. From [this post](https://developers.google.com/web/fundamentals/performance/rendering/stick-to-compositor-only-properties-and-manage-layer-count):

> The problem here is that every layer you create requires memory and management, and that’s not free.

>Every layer’s textures needs to be uploaded to the GPU, so there are further constraints in terms of bandwidth between CPU and GPU, and memory available for textures on the GPU.

Some random thoughts that would still need testing:

- The blinking cursor of the search input needs constant repainting which doesn't play well when trying to have the whole panel as a compositor layer. Maybe we could not auto-focus the search input so there wouldn't be any blinking?
- Or we could have the top part with the search input fixed and only the list with the results (and favorites) would scroll? Then auto-focus might be fine.
- Only add `will-change: transform;` when hovering that list and not while typing a search. Hovering the list doesn't mean you want to scroll but maybe still better than keep it on all the time.
- Remove the hover state of the packages. I think it's fine when you scroll fast, but when you scroll slowly it constantly has to repaint the package card background.
- It also seems that removing the scrollbars `.settings-view .panels .panels-item::-webkit-scrollbar { display: none; }` makes scrolling smoother, but then usability suffers.

I'll test a bit more and also take a look the tree-view.

simurai:
Some more paint testing with the timeline:

Below the recordings when the mouse doesn't touch any cards and was just on the side when scrolling.

No layer | No layer - long list | With layer - long list 
--- | --- | ---
no-layer | no-layer-long-list | layer-long-list
The whole window gets painted (`1000x800`) | When the list gets longer, like when searching for `linter`, it still just paints the window dimensions (`1000x800`) | With compositor layer, the texture depends on the list size, in this case (`548x5028`)

This would suggest:

- Don't use a layer when the window is small and the list is long.
- Use a layer if the window is large and the list is short.

But I'm not even sure if the size of the painting area matters? It seems the frame rate is similar in both cases (`15-20ms`). Also, I guess it depends which hardware is used and probably other things.

When having the mouse on top of cards (with the hover effect of the background changing in color) and scrolling slowly, it seems to make a big difference.

No layer | With layer
--- | ---
no-layer-long-list-hover | layer-long-list-hover

Frame rate is much worse with a layer. I guess it's because the GPU needs to be constantly updated with the changed background color of the cards.

easyaspi314:
@simurai Well in my opinion, until we get it fixed, we should let the GPU do the work. 

Decreased frame rate is less of an issue than heavy CPU usage, for now. 

Also, I saw [Chromium issue 317007](https://bugs.chromium.org/p/chromium/issues/detail?id=317007), titled ":hover effects cause hard-to-prevent repaint storms". Which is DEFINITELY our problem.

I suggest everyone go star that. 

However, looking at that, maybe we can put `pointer-events: none` whenever we scroll. Would that work?

I can't test at the moment.

simurai:
Yeah, `pointer-events: none` helped, but now I can't reproduce the slower frame rate even without:

screen shot 2016-12-15 at 3 00 35 pm

Much less red. Maybe there was something else going on at the same time. 🤔 

@easyaspi314 Any concerns on other potential side effects. Same quote from above taken from [this post](https://developers.google.com/web/fundamentals/performance/rendering/stick-to-compositor-only-properties-and-manage-layer-count):

> The problem here is that every layer you create requires memory and management, and that’s not free.

>Every layer’s textures needs to be uploaded to the GPU, so there are further constraints in terms of bandwidth between CPU and GPU, and memory available for textures on the GPU.

I currently have this in my `styles.less` file, so I'll test it a bit to see if I notice anything unusual:

```less
// PR https://github.com/atom/tree-view/pull/958
.tree-view-scroller {
  will-change: transform;
}

// PR https://github.com/atom/settings-view/pull/867
.settings-view {
  will-change: transform;

  .config-menu {
    will-change: transform;
    .panels-packages{
      will-change: transform;
    }
  }
  .panels {
   .panels-item {
      will-change: transform;
    }
  }
}

// PR https://github.com/atom/atom-space-pen-views/pull/20
.select-list {
  ol.list-group {
    will-change: transform;
  }
}
```

damieng:
Any further discussion on this?

simurai:
Maybe @as-cii has some feedback about using GPU layers in the settings view?

I tend towards a `wontfix`. Just because the impact is hard to test on all the different platforms and people's machines.

as-cii:
Apologies for the delay in answering this, @easyaspi314! And thanks for your contribution. ⚡️ 

As far as I understand from the measurements posted so far this does not look like a clear win, so I would lean toward not merging it to avoid issues like dropping sub-pixel anti-aliasing support.

However, I will be happy to reconsider this if/when there will be benchmarks that clearly indicate this pull request is a net improvement over the current situation. 👍

PR #872, opened 2016-10-23
Hide unnamed grammars in package-detail view

Alhadis:
It isn't uncommon for language packages to include a few "helper" grammars for internal use. Most of the time, this is to provide embeddable highlighting for one or more "real" grammars. These shouldn't be visible to the end user... and certainly not with a title of `undefined Grammar`:

Figure 1

**Complements:** atom/grammar-selector#34

### DIFF
--- a/lib/package-grammars-view.coffee
+++ b/lib/package-grammars-view.coffee
@@ -46,11 +46,11 @@ class PackageGrammarsView extends View
     @grammarSettings.empty()
 
     for grammar in @getPackageGrammars()
-      {scopeName} = grammar
-      continue unless scopeName
+      {scopeName, name} = grammar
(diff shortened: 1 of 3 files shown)


Alhadis:
Sigh... Travis brings nothing but pain, I swear.

Figure 2

Specs pass for me...

Alhadis:
@thomasjo Any idea why the specs are passing 4/5 times locally, but failing every time in CI?

Can you reproduce this locally? Even on master?

damieng:
The test that's failing is flakey in master already and isn't caused by your code.

Thanks for your contribution!

Alhadis:
No problem! :-)

I thought it was, because it failed on master too. Thanks!

PR #884, opened 2016-12-07
Indicate when a package is updating in the status bar tile

### LINKED ISSUE #883: Updates status bar should indicate when a package is updating
This is a natural follow-up to #778 and #511.

When packages are updating, the status bar tooltip should reflect that to give active feedback.  I'm thinking something like `5 updates (3 updating)` if there are a total of 5 updates available, and 3 packages are currently updating.  If all the packages are currently updating, then the tooltip could become `4 updating`.  If an update fails, perhaps that can also be displayed: `1 update failed`.

I am going to tentatively assign myself to this, but I may not get to it in a reasonable amount of time.
Therefore, if anyone is reading this and would like to attempt it, please feel free to do so.

50Wliu:
Closes #883

Some things to note:
* The tile changes have been extracted into a helper `updateTile` method which I think helps streamline and deduplicate things (see below).
* This PR fixes a bug in my previous PR where I accidentally set the priority of the tile to 0 when reinitializing it instead of -99.  The new helper method takes care of this.
* This PR also fixes a different bug in my previous PR where we were trying to call `tooltip.dispose()` on a non-existent tooltip if the Check for Updates button was pressed.  Whoops!
* I fixed a bug in PackageManager where the `package-updating` event was being generically emitted through Emitter rather than through the dedicated `emitPackageEvent` helper method.  This means that `theme-updating` will now start to be emitted for themes.

TODO:
* [x] Finish what to display for failed updates
* [ ] Get some feedback on the label text
* [x] Specs

### DIFF
--- a/lib/package-manager.coffee
+++ b/lib/package-manager.coffee
@@ -280,7 +280,7 @@ class PackageManager
         error.stderr = stderr
         onError(error)
 
-    @emitter.emit('package-updating', {pack})
+    @emitPackageEvent 'updating', pack
     apmProcess = @runCommand(args, exit)
     handleProcessErrors(apmProcess, errorMessage, onError)
 
--- a/lib/package-updates-status-view.coffee
+++ b/lib/package-updates-status-view.coffee
@@ -44,7 +44,7 @@ class PackageUpdatesStatusView extends View
       if update.name is pack.name
         @updatingPackages.splice(index, 1)
 
-    for index, update in @failedUpdates
+    for index, update of @failedUpdates
       if update.name is pack.name
         @failedUpdates.splice(index, 1)
 
--- a/lib/package-updates-status-view.coffee
+++ b/lib/package-updates-status-view.coffee
@@ -13,26 +13,15 @@ class PackageUpdatesStatusView extends View
     @updatingPackages = []
     @failedUpdates = []
 
-    packageManager.on 'package-updated theme-updated', ({error, pack}) => @onDidUpdatePackage(pack)
     packageManager.on 'package-update-available theme-update-available', ({error, pack}) => @onPackageUpdateAvailable(pack)
     packageManager.on 'package-updating theme-updating', ({error, pack}) => @onPackageUpdating(pack)
+    packageManager.on 'package-updated theme-updated', ({error, pack}) => @onPackageUpdated(pack)
     packageManager.on 'package-update-failed theme-update-failed', ({error, pack}) => @onPackageUpdateFailed(pack)
 
     @updateTile()
 
     @on 'click', ->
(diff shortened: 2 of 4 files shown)


50Wliu:
Thoughts: should the `x updates` text include the number of packages that are currently updating?  For example, if there are 5 packages and 3 are updating, should the text read `5 updates (3 updating)`, or `2 updates (3 updating)`?  Or maybe `3/5 updating`, `3 out of 5 updating`?

I think I like the last two the most.

In addition, regarding failed updates, should we remove the failed text when the tile is clicked?  And should we remove that package from the available updates count?

/cc @atom/feedback

damieng:
I'd like "5 package updates available"  and once the process starts "5 packages updating" then "4 packages updating"... until gone.  i.e.  instead of 3-5 just count down from 5.

simurai:
> @damieng once the process starts "5 packages updating"

I like that too because it's simpler. But what if there a 5 packages in total and you only click the update button for 3 packages. In that case "5 packages updating" sounds wrong.

Hmm.. maybe still write it out? It's longer but also clearer. The above example would be:

- `5 package updates available`
- `5 package updates available (3 updating)`
- `4 package updates available (2 updating)`
- `3 package updates available (1 updating)`
- `2 package updates available`

> In addition, regarding failed updates, should we remove the failed text when the tile is clicked? And should we remove that package from the available updates count?

Maybe keep it as long as the settings-view shows an update as failed. I guess if you reload Atom it will show again as "normal" update.

- `5 package updates available`
- `5 package updates available (3 updating)`
- `4 package updates available (2 updating)`
- `4 package updates available (1 updating, 1 failed)`
- `3 package updates available (1 failed)`

Actually, is this only for the tooltip? Or will the status-bar tile also show if updates are in progress?

50Wliu:
This is just for the text displayed in the status bar.  The tooltip is more verbose and is already basically what you're suggesting: `5 package updates available (3 currently updating)`.

simurai:
Ok, for the status-bar I like your `3/5 updating` suggestion. Same example from [above](https://github.com/atom/settings-view/pull/884#issuecomment-265637896) 5 total only 3 get clicked:

- `5 updates`
- `3/5 updating`
- `2/4 updating`
- `1/3 updating`
- `2 updates`

And when failing:

- `5 updates`
- `3/5 updating`
- `2/4 updating`
- `1/4 updating (1 failed)`
- `3 updates (1 failed)`

50Wliu:
Ok I think this is ready.  The only thing that might still have room for improvement is the failed text.

lee-dohm:
It looks good to me 👍 It appears that you've gone with @simurai's suggestions on the tooltips and status bar text, so that sounds like the best solution for now. We can always come back and tweak it if things go awry.

PR #887, opened 2016-12-08
Drastically improve loading time of updates panel

### LINKED ISSUE #409: Clicking the 'package updates available' icon searches again for package updates
When Atom is started, it looks for package updates, and if it finds some an icon is displayed in the status bar indicating I should update my packages. However, when I click that button Atom searches <em>again</em> for package updates instead of using the information it got just some seconds ago... This takes quite some time, so I try to get some work done in the meantime. But now I forgot that there were package updates in the first place...

50Wliu:
`getInstalled()` doesn't affect `getOutdated()` and was just slowing the process down if we had cached data from `getOutdated()`.

Here's two GIFs, the first one before the change and the second one after.  Pretty big difference :).  Sorry for the extra non-Atom stuff in the background.  Had to resize the Atom window but forgot to resize licecap.
Before:
wl-updates-panel-before

After:
wl-updates-panel-after

Fixes #409, closes atom/atom#6945.
Though this doesn't do anything about the case when you click the updates tile a sufficiently long time after Atom loads, causing Atom to check for updates again since the outdated cache expired.  I think that's fine though.

/cc @Caged

### DIFF
--- a/lib/updates-panel.coffee
+++ b/lib/updates-panel.coffee
@@ -52,15 +52,14 @@ class UpdatesPanel extends ScrollView
 
     @checkingMessage.show()
 
-    @packageManager.getInstalled().then =>
-      @packageManager.getOutdated()
-        .then (@availableUpdates) =>
-          @checkButton.prop('disabled', false)
-          @addUpdateViews()
-        .catch (error) =>
-          @checkButton.prop('disabled', false)
-          @checkingMessage.hide()
-          @updateErrors.append(new ErrorView(@packageManager, error))
+    @packageManager.getOutdated()
+      .then (@availableUpdates) =>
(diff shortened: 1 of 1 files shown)


simurai:
Wow.. that looks so much faster! 🏇 👏

Caged:
Niiiiice.  👏

50Wliu:
Merging.  I don't think a spec needs to be added for this (what would it even test?).

eboracus:
Will this effect the delay on loading the cards in the `packages` tab aswell? Does that use the same `getInstalled()`?

50Wliu:
No.  This only affects the updates panel.  I plan on taking a look at that soon though!  I noticed we're caching the packages, but then not using it in most cases.

PR #893, opened 2016-12-13
[WIP] Fix crazy metadata chaos

### LINKED ISSUE #814: [Enhancement] Update button on detail view of package
Replicate the "Update" button in the Available Updates view that is on the list of each package, to also be in the detail view of that package.

**Current and suggested behavior**

I often find myself seeing that there is packages to update in the "Updates" page.  Then click into the package to read the changelog/description of that package and then from in there, there is no update button.  You then have to go back to the "Updates" view to actually update the package.

**Why would the enhancement be useful to most users**

If you're like me it saves you one whole click (😱) so that you don't have to go back to the list of available updates.

**Screenshot**

Example

**Atom Version:** 1.90-beta0
**OS and Version:** macOS Sierra (10.12 beta 2)

50Wliu:
Metadata was being included in what was already metadata, package detail
view was loading a new package card despite the fact that package card
calls the detail view, incorrect metadata was being loaded in certain
cases, activateConfig isn't even needed, and more.

Fixes #814 and probably more - will test later.

This still has a few bugs associated with it, but all specs pass...

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,40 @@
+<!--
+
+Have you read Atom's Code of Conduct? By filing an Issue, you are expected to comply with it, including treating everyone with respect: https://github.com/atom/atom/blob/master/CODE_OF_CONDUCT.md
+
+Do you want to ask a question? Are you looking for support? The Atom message board is the best place for getting support: https://discuss.atom.io
+
+-->
+
+### Prerequisites
+
+* [ ] Put an X between the brackets on this line if you have done all of the following:
+    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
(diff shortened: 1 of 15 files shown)


50Wliu:
Not sure if it's just me or caused by this PR, but I'm getting tons of `Uncaught (in promise) DOMException: Failed to execute 'put' on 'IDBObjectStore': An object could not be cloned.` errors whenever I enter a package detail view.  It also prevents me from reloading Atom.

50Wliu:
Ok.  So there's two options here:
1. Add `loadTime`/`activateTime` to the package's metadata (atom/atom#13438).  I'm not sure if this is consistent with what the metadata should contain.  In addition, use `atom.packages.resolvePackagePath(name)` for changelog/license/readme purposes.
2. Extend the package metadata with the actual Package object from `atom.packages.getLoadedPackage(name)`.  This is ugly because we end up duplicating the metadata.

50Wliu:
Will close and reopen against latest master to avoid dealing with conflicts.

PR #894, opened 2016-12-16
Use atom.packages.getAvailablePackageNames

### LINKED ISSUE #817: "Install" button is shown for disabled package after reloading detail view
image

Reproduced with 0.240.1
### Steps to reproduce
1. Show Packages
2. Select single package to show detail
3. Disable the package (if enabled)
4. Run window:reload command
### Behavior
- Expected: "Enable" button is shown
- Actual: "Install" button is shown

50Wliu:
Because it's already there for us!  Why use our own implementation?

Fixes #817

TODO:
* [x] Specs

### DIFF
--- a/lib/package-manager.coffee
+++ b/lib/package-manager.coffee
@@ -11,7 +11,6 @@ class PackageManager
 
   constructor: ->
     @packagePromises = []
-    @availablePackageCache = null
     @apmCache =
       loadOutdated:
         value: null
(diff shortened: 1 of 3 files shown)


50Wliu:
Turns out there was already a spec testing for this behavior!  The only problem was that the assertion was wrong...

50Wliu:
Feeling pretty confident about this change (minimally invasive, uses core APIs that rely on the filesystem), so I'm going to merge it.  This might be a bit slower than the current implementation since it hits the filesystem every time but it should be fine.  My (not-very-scientific) tests don't show any noticeable slowdown.

PR #898, opened 2017-01-12
Apply Chromium proxies to env vars for apm

damieng:
### Description of the Change

Chromium does a good job at reading system proxy settings which means a lot of what Atom needs to do goes through the correct proxies.

The apm command line being a regular node app and not an Electron/Chromium app does not read or have access to these proxy settings and so fails when trying to connect if you are using a proxy.

This change determines what http and https proxy servers to use based on Chromium's session.resolveProxy call with atom.io and then sets the http_proxy and https_proxy environment variables that apm then uses.

### Alternate Designs

Alternative designs include;

1. Moving apm functionality entirely into Atom - too much work at this time and would mess up command-line scripting
2. Editing the apm.rc file with the proxy settings - too destructive

### Benefits

People behind proxy servers can use apm with no effort.

### Possible Drawbacks

Users with manually set http_proxy and https_proxy environment variables will find we override their preferences. They can switch off the auto-detect proxy in settings.

### DIFF
--- a/lib/package-manager.coffee
+++ b/lib/package-manager.coffee
@@ -37,6 +37,21 @@ class PackageManager
     schema = atom.config.getSchema(packageName)
     schema? and (schema.type isnt 'any')
 
+  setProxyServers: (callback) =>
+    session = atom.getCurrentWindow().webContents.session
+    session.resolveProxy 'http://atom.io', (httpProxy) =>
+      @applyProxyToEnv 'http_proxy', httpProxy, =>
+        session.resolveProxy 'https://atom.io', (httpsProxy) =>
+          @applyProxyToEnv 'https_proxy', httpsProxy, callback
+
+  applyProxyToEnv: (envName, proxy, callback) ->
+    if proxy?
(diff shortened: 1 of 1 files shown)


lib/package-manager.coffee
```diff
@@ -37,6 +37,22 @@ class PackageManager
     schema = atom.config.getSchema(packageName)
     schema? and (schema.type isnt 'any')
 
+  setProxyServers: (callback) =>
+    callback() if atom.config.get('core.useProxySettingsWhenCallingApm') isnt true
```
50Wliu:
```coffee
callback() unless atom.config.get('core.useProxySettingsWhenCallingApm')
```
seems cleaner.

50Wliu:
Also, this looks like it only gets used with `runCommand`, so maybe `setProxyServers` can be moved to inside `runCommand`?

damieng:
I originally had it in runCommand but the problem is that runCommand immediately passed back a BufferedProcess object that things attach to - this being async it couldn't do that immediately. Breaks everywhere.

PR #916, opened 2017-02-23
Show errors in InstalledPackagesPanel and ThemesPanel

50Wliu:
### Requirements

### Description of the Change

From the looks of it, InstalledPackagesPanel and ThemesPanel were mostly a result of copy-paste from UpdatesPanel and InstallPanel.  And some of the variables forgot to be renamed...resulting in a hidden uncaught exception when attempting to display the error.  Errorception.
So: add an extra div that wraps around all the sub-package listings (community, core, etc).  When something goes :boom:, hide that div and also the total package count, then display the error.  The error div itself has been renamed from `updateErrors` and `themeErrors` to just `errors`.

### Alternate Designs

This is primarily a bugfix, so no alternate designs were considered.

### Benefits

No more "hmm, the packages listing is taking a _really_ long time to load..." 🤔

### Possible Drawbacks

None

### Applicable Issues

None

Whitespace-insensitive diff: https://github.com/atom/settings-view/pull/916/files?w=1

### DIFF
--- a/lib/installed-packages-panel.coffee
+++ b/lib/installed-packages-panel.coffee
@@ -82,7 +82,7 @@ class InstalledPackagesPanel extends CollapsibleSectionPanel
 
     @packageManagerSubscriptions = new CompositeDisposable
     @packageManagerSubscriptions.add @packageManager.on 'package-install-failed theme-install-failed package-uninstall-failed theme-uninstall-failed package-update-failed theme-update-failed', ({pack, error}) =>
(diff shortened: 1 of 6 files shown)


50Wliu:
Closing and reopening under latest master to avoid dealing with these conflicts.

50Wliu:
Actually, looks like this was fixed in the jQuery removal PR! :heart:

PR #918, opened 2017-02-25
Make update buttons wrap properly

tomsapps:
### Description of the Change
This allows the action buttons in the updates-panel header wrap properly on smaller screen sizes.
(see images below)

### Alternate Designs
None just stayed consistent with the current ui.

### Benefits
Clean & functional ui on smaller screens

### Possible Drawbacks
None

### Applicable Issues
None

current_wrap

both_action_btns


tomsapps:
@simurai nice! that definitely cleaned it up! I agree with you on giving themes the ability to override, did not think about that :P

PR #923, opened 2017-03-15
Wait for `open()` to complete before considering command done

matthewwithanm:
This has been relying on the pane to be created synchronously when
dispatching the command (and calling `open()`), which would only happen
in certain conditions. After atom/atom#13977, `open()` is more reliably
async so we need to take that into account.


maxbrunsfeld:
Thanks for tracking this down. I made a slight tweak to avoid using a spy.

PR #929, opened 2017-03-28
Use `atom.packages.resolvePackagePath` instead of `pack.path`

as-cii:
Bundled packages will be preloaded during `script/build` and, since Atom cannot provide an absolute path during the snapshot phase, it will assign a relative path (`node_modules/{packageName}`) to preloaded packages. 

With this pull request we will resolve such path at runtime via `atom.packages.resolvePackagePath` in order to load e.g. readmes when opening a package card.


as-cii:
Closing this, as we can turn the relative path into an absolute path right before activating packages at runtime with basically no additional cost.

PR #950, opened 2017-05-15
Prevent version pinned packages from showing as needing updates

lee-dohm:
### Description of the Change

Prevent package whose names are listed in `core.versionPinnedPackages` from being shown for automatic updating. This is only prevented through the Atom UI. Using `apm` from the command line will not respect the `core.versionPinnedPackages` setting.

### Alternate Designs

* Specify the version number in addition to the package name
    * Decided that simply specifying the package name simplified the code and was probably good enough for the target use case
    * If people want to update to a newer version, they would need to update the package _and_ update their config file
* Prevent the packages from being updated within `apm` as well
    * Figured that people should be able to easily update their pinned version if they wanted but keep the package pinned

### Benefits

People can prevent Atom from recommending to update to a newer version of specific packages when the newer version has bugs or problems.

### Possible Drawbacks

People could get "stuck" on old versions of packages that have critical improvements. (Added alert to the updates panel to help with this.)

### Applicable Issues

* atom/atom#14417

### DIFF
--- a/lib/package-manager.coffee
+++ b/lib/package-manager.coffee
@@ -146,8 +146,7 @@ class PackageManager
           error = createJsonParseError(errorMessage, parseError, stdout)
           return callback(error)
 
-        pinnedPackages = atom.config.get('core.versionPinnedPackages') ? []
-        updatablePackages = (pack for pack in packages when not pinnedPackages.includes(pack?.name))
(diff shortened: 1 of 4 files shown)


lee-dohm:
@50Wliu Added a warning alert on the updates panel for when there are version pinned packages:

screen shot 2017-05-15 at 12 37 47 pm

PR #954, opened 2017-05-24
Hidden settings

smashwilson:
### Requirements

### Description of the Change

Support a `hidden` attribute in package configuration schema. If present and set to something truthy, the setting is not rendered in the Settings View, but can only be interacted with manually through `atom.config.set()` or by editing `config.cson`.

### Alternate Designs

I considered blacklisting a package name like `hidden` on atom.io and using `atom.config.set('hidden.someSetting')` as a quicker way to achieve non-user-facing options, but I wasn't fond of that because it creates a single, global namespace for hidden settings, and because it would cause additional confusion for anyone who looked at their raw `config.cson`.

I also considered using a visible "show on load" setting like we have in the welcome package, but we don't have an obvious place to put the checkbox on our panels, and I didn't want to make users hunt for it in the settings view. If we used a visible setting and automatically toggled it off after the first reveal, that would cause even _more_ confusion because we'd be messing with something that they explicitly enabled.

### Benefits

In the GitHub package, we want to show the Git and GitHub dock items on launch to aid discoverability, but only on the first Atom run. We're currently using state serialization, but that has the drawback of being scoped to the current project, so the dock items reappear each time you open a new project.

Hidden settings provide a mechanism for storing Atom-global package state, which we can use to store a latch setting that tracks whether or not we should launch with the panels present or not.

### Possible Drawbacks

It introduces a disparity between the configuration seen in the settings view and the one seen in your config file, so I wouldn't overuse it. If the config file gets corrupted somehow it would be harder for users to diagnose the problem.

/cc @BinaryMuse @kuychaco @nathansobo @lee-dohm

### DIFF
--- a/lib/settings-panel.js
+++ b/lib/settings-panel.js
@@ -344,6 +344,11 @@ function elementForSetting (namespace, name, value) {
     }
   }
 
+  let schema = atom.config.getSchema(`${namespace}.${name}`)
+  if (schema && schema.hidden) {
+    return document.createDocumentFragment()
+  }
+
   const controlGroup = document.createElement('div')
   controlGroup.classList.add('control-group')
 
@@ -351,7 +356,6 @@ function elementForSetting (namespace, name, value) {
   controls.classList.add('controls')
   controlGroup.appendChild(controls)
 
-  let schema = atom.config.getSchema(`${namespace}.${name}`)
   if (schema && schema.enum) {
     controls.appendChild(elementForOptions(namespace, name, value))
   } else if (schema && schema.type === 'color') {
--- a/spec/settings-panel-spec.coffee
+++ b/spec/settings-panel-spec.coffee
@@ -183,3 +183,29 @@ describe "SettingsPanel", ->
       expect(controlGroups[1].querySelector('.sub-section .sub-section-heading').classList.contains('has-items')).toBe true
       # Should be already collapsed
       expect(controlGroups[1].querySelector('.sub-section .sub-section-heading').parentElement.classList.contains('collapsed')).toBe true
+
+  describe 'hidden settings', ->
+    beforeEach ->
+      config =
+        type: 'object'
+        properties:
+          visible:
+            name: 'visible'
+            title: 'A visible setting'
+            description: 'This setting should appear in the settings view'
+            type: 'string'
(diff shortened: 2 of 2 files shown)


damieng:
I'd much rather have the settings not hidden. If I launched Atom to do
something in a hurry and ignored the GitHub package stuff I'd be
disappointed there's no way to turn it back on without digging into config
files.

Elsewhere in Welcome and Metrics we specifically have a checkbox or choice
to get rid of it...

simurai:
What about adding a new button/section to the Welcome Guide instead of having the Git panel opened on first run?

As much as I think we should promote the Git panel, seeing it right there feels a bit too early and distractive:

screen shot 2017-05-25 at 12 26 32 pm

Also, if people miss it in the guide, there is still the status-bar button that people probably click at some point.

screen shot 2017-05-25 at 12 34 28 pm

lee-dohm:
Given that all of the other settings currently in the GitHub package are debug or diagnostic related except one, I don't believe this needs to be hidden, especially since the analogous ones in Welcome and Metrics aren't.

The traditional way to hide configuration completely from the Settings View is simply to not save it in the `config.cson`. Other packages (including my own [package-sync package](https://atom.io/packages/package-sync)) store separate configuration in the `~/.atom` directory next to the `config.cson` and other Atom configuration files. This also maintains the contract of our instructions for [resetting Atom to factory defaults](http://flight-manual.atom.io/hacking-atom/sections/debugging/#reset-to-factory-defaults).

smashwilson:
> Given that all of the other settings currently in the GitHub package are debug or diagnostic related except one, I don't believe this needs to be hidden, especially since the analogous ones in Welcome and Metrics aren't.

This one is slightly different in that the package itself forces it to `false` on activation. I'd feel weird showing a setting that the package was always going to force off the next time it started.

> The traditional way to hide configuration completely from the Settings View is simply to not save it in the config.cson. Other packages (including my own package-sync package) store separate configuration in the ~/.atom directory next to the config.cson and other Atom configuration files. This also maintains the contract of our instructions for resetting Atom to factory defaults.

Oh, interesting. 💭 Yeah, that's a better alternative, I think. Thanks 🙇

PR #958, opened 2017-06-01
prevent disabled packages from triggering status view updates

### LINKED ISSUE #956: Update status message should not count disabled packages
Moved from https://github.com/atom/settings-view/issues/955

> Disabled packages should not be counted (`(N) Updates` message at the bottom-right corner)

zoo1:
### Description of the Change

Check when a package update is available if it is disabled and prevent a status view alert.
Fixes #956 

### Benefits

Remove unnecessary update noise to users.

### DIFF
--- a/lib/package-updates-status-view.js
+++ b/lib/package-updates-status-view.js
@@ -53,6 +53,10 @@ export default class PackageUpdatesStatusView {
   }
 
   onPackageUpdateAvailable (pack) {
+    if (atom.packages.isPackageDisabled(pack.name)) {
+      return
+    }
+
     for (const update of this.updates) {
       if (update.name === pack.name) {
         return


50Wliu:
Looks like I labeled that as a bug a bit too hastily.  Looking back, I think this is expected behavior: there _is_ still an update available for the disabled package, and we should probably be displaying that in the status view.

@KSXGitHub is there any reason why you think disabled packages shouldn't be displayed in the status bar?

KSXGitHub:
@50Wliu There is no reason I would want to frequently update a disabled package because I do not use it frequently. If I really want to update a disabled package, I can just go to settings view or just use apm.

> Looking back, I think this is expected behavior

I think you should label #956 as a feature rather than a bug

50Wliu:
Ok, I now agree with this change.  However, there needs to be additional logic in `onPackageUpdating` and `onPackageUpdateFailed` to add the disabled package to `this.updates`.  Either that, or adding the disabled check to all the `on*` methods.  Also, specs will be needed.

PR #963, opened 2017-06-29
Display keybindings even when disabled

### LINKED ISSUE #946: Disabling keybindings removes the keybindings section and toggle
### Prerequisites

### Description

When disabling package keybindings, the list of keybindings disappears.  Furthermore, after restarting Atom, the entire keybindings section along with the keybinding toggle disappears.  This means that you have to go into your `config.cson` to re-enable the keybindings.

### Steps to Reproduce

1. Find a package that has keybindings and open its details page in the Settings View
2. Uncheck the "Enable" box under Keybindings
3. Restart Atom

**Expected behavior:** After step 2: for the keybindings to still be visible.  After step 3: for the keybindings and toggle checkbox to still be there.

**Actual behavior:** After step 2: the keybindings list disappears.  After step 3: the entire keybindings section disappears.

**Reproduces how often:** 100%

### Versions

```
Atom    : 1.18.0-dev-625a2153c
Electron: 1.6.5
Chrome  : 56.0.2924.87
Node    : 7.4.0
```

### Additional Information

Thanks to @TevinLi for the initial report.
/cc @as-cii for possible regression from jQuery removal.

Here's some pictures copied from the original issue.
image
image
image

smashwilson:
Collect the keybindings created by each package from the package itself rather than filtering them from the `atom.keymaps` global. `atom.keymaps` does not include bindings from disabled packages or from packages with disabled keybindings, but the package's `.keymaps` property does.

As a side effect, we'll no longer be relying on the convention of beginning command names with the package name to identify which commands are associated with a package. Weird Things will happen if you deviate from this with the present scheme!

Fixes #946. Depends on atom/atom-keymap#218 for `atom.keymaps.build()`.

### DIFF
--- a/lib/package-keymap-view.js
+++ b/lib/package-keymap-view.js
@@ -114,7 +114,7 @@ export default class PackageKeymapView {
       // Backwards compatibility for Atom <= 1.19
       for (const keyBinding of atom.keymaps.getKeyBindings()) {
         const {command} = keyBinding
-        if (command && command.indexOf && command.indexOf(`${this.namespace}:` === 0)) {
+        if (command && command.indexOf && command.indexOf(`${this.namespace}:`) === 0) {
           keyBindings.push(keyBinding)
         }
       }
--- a/lib/package-keymap-view.js
+++ b/lib/package-keymap-view.js
@@ -114,7 +114,7 @@ export default class PackageKeymapView {
       // Backwards compatibility for Atom <= 1.19
       for (const keyBinding of atom.keymaps.getKeyBindings()) {
         const {command} = keyBinding
-        if (command && command.indexOf && command.indexOf(`${this.namespace}` === 0)) {
+        if (command && command.indexOf && command.indexOf(`${this.namespace}:` === 0)) {
           keyBindings.push(keyBinding)
         }
       }
--- a/lib/package-keymap-view.js
+++ b/lib/package-keymap-view.js
@@ -106,8 +106,18 @@ export default class PackageKeymapView {
     }
 
     const keyBindings = []
-    for (const [keymapPath, keymap] of atom.packages.getLoadedPackage(this.namespace).keymaps) {
-      keyBindings.push(...atom.keymaps.build(this.namespace, keymap, 0, false))
+    if (atom.keymaps.build) {
+      for (const [keymapPath, keymap] of atom.packages.getLoadedPackage(this.namespace).keymaps) {
+        keyBindings.push(...atom.keymaps.build(this.namespace, keymap, 0, false))
+      }
+    } else {
+      // Backwards compatibility for Atom <= 1.19
+      for (const keyBinding of atom.keymaps.getKeyBindings()) {
+        const {command} = keyBinding
+        if (command && command.indexOf && command.indexOf(`${this.namespace}` === 0)) {
+          keyBindings.push(keyBinding)
+        }
+      }
     }
 
     for (const keyBinding of keyBindings) {
--- a/lib/package-keymap-view.js
+++ b/lib/package-keymap-view.js
@@ -97,8 +97,10 @@ export default class PackageKeymapView {
   updateKeyBindingView () {
     this.refs.keybindingItems.innerHTML = ''
 
-    const keymaps = atom.packages.getLoadedPackage(this.namespace).keymaps
-    const keyBindings = atom.keymaps.build(this.namespace, keymaps, 0, false)
+    const keyBindings = []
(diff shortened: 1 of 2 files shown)


smashwilson:
It's working: 🎊 

keybindings

I tested a reload with the keybindings disabled and that works now, too.

A few things I'm mulling over, which I'm curious to collect others' thoughts on:

1. One thing you'll notice here is that this does include _all_ keybindings registered by any keymaps in the package... even the `core:*` events for dialogs and such. I'm leaning toward counting that as a feature rather than a bug. After all, those are the exact set of bindings that the checkbox will enable or disable.

2. I was considering adding `.text-subtle` to the keybinding table rows to grey them out when they're disabled. Good idea?

3. This still doesn't display keybindings when the entire package is disabled, I'm assuming because its keymaps aren't loaded at all. We don't parse or display any of the settings from a package while it's disabled, as it turns out, I'm assuming because "disabled" == "nothing parsed from its metadata at all". That's probably a good idea, honestly, so unless anyone thinks otherwise I'll likely keep it that way.

/cc @maxbrunsfeld @50Wliu @as-cii for comments and feels

smashwilson:
Oh, also: the tests are going to be failing until atom/atom-keymap#218 lands in stable. Is there a protocol for dealing with that? Should I make the effort to fail back to the old behavior if `atom.keymaps.build()` doesn't exist, and skip the new assertions?

I feel like we'd ultimately want to test core packages against Atom master, or at least a nightly 🤔

as-cii:
> Should I make the effort to fail back to the old behavior if atom.keymaps.build() doesn't exist, and skip the new assertions?

Yeah, this is a policy we usually try to enforce in order to have green tests but I agree it's suboptimal, especially because it doesn't really give us much confidence over the added behavior (e.g. does it work against master, beta or stable?) and we have to wait until bumping the version on atom/atom to know whether or not a particular change broke something.

> I feel like we'd ultimately want to test core packages against Atom master, or at least a nightly 🤔

While working on something else, @jasonrudolph and I have whipped up a [script](https://gist.github.com/as-cii/70344191e8618de74593d97b90c758f1) that allows packages to be built against the latest green master build. It's a slightly modified version of the script that we ship in [atom/ci](https://github.com/atom/ci) and it uses a new simple script called `last-atom-master-artifact-url.js` to detect the latest green build and retrieve the associated macOS artifact URL.

Moving forward I think it might be interesting to experiment with a similar approach, so I wonder if we could incorporate the above two scripts in [atom/ci](https://github.com/atom/ci) (making it work also on Windows/AppVeyor and Linux/Travis), and provide a new switch to target them (`RELEASE_CHANNEL=master`) in addition to the existing `stable` and `beta`. 

@smashwilson (cc: @atom/maintainers): what do you think about trying it out on this package? The only downside I can think of is that initially we would have an inconsistent CI setup, where some packages target stable/beta and others target master.

smashwilson:
> what do you think about trying it out on this package?

:+1: :sparkles: I'll see what I can whip up and use this PR as a guinea pig.

For core packages, I'm thinking that we really want a build matrix of beta, stable, and master, with beta and stable flagged as "allowed failures." Then I could modify the [build-status matrix](https://github.com/atom/atom/blob/master/docs/build-instructions/build-status.md) to give us a bird's eye view of whether or not each package is green with respect to each channel, which would tell us whether or not it's safe to :arrow_up: to the current version for hotfixes against each.

smashwilson:
Also, it wouldn't be _too_ messy to modify this to work against beta/stable. I just hesitated to because I know I'm not going to remember to clean it up after we roll the railcars 😆

smashwilson:
Well, I started to have a look at implementing "build against master" in atom/ci, but it isn't immediately obvious to me what the best way to implement it is for non-Circle artifacts. I have a few different approaches I was evaluating yesterday but I think any of them would be > 1 day of work, and I'm thinking it'd be better to not hold up the reactor queue for them. @as-cii, how would you feel about moving that discussion to a different issue and shipping this as-is? I'm guessing we won't have a shortage of core package PRs to try it out on :grin:

> Also, it wouldn't be too messy to modify this to work against beta/stable.

I went ahead and did this to get us green builds against beta and stable anyway.

@damieng, care to give this a look-over and see if I missed anything?

as-cii:
> how would you feel about moving that discussion to a different issue and shipping this as-is? I'm guessing we won't have a shortage of core package PRs to try it out on 😁

Agreed. 👍

damieng:
Tested this on my box and it works as expected, solves the issues nicely against master.

smashwilson:
> Tested this on my box and it works as expected, solves the issues nicely against master.

Excellent, thanks :sparkles: Shipping presently.

Issue #987, opened 2017-09-01
Relative image links in READMEs don't work


frasertmay:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

When viewing a README for a package, images using relative links are not displayed as Atom can't find the files. For example if the package keeps images in `<git repo>/images/blahblah.jpg` and the README says `![Image](images/blahblah.jpg)` - viewing the README on GitHub and on Atom.io the images appear correctly however on Atom they are not found.

![screen shot 2017-09-01 at 18 28 16](https://user-images.githubusercontent.com/3003251/29981084-4e9414e2-8f44-11e7-8f24-2487136f2c15.png)

### Steps to Reproduce

1. Search for a package which uses relative links for images in its README (such as [atom-clock](https://atom.io/packages/atom-clock))
2. Look at the README  and see that they don't resolve

**Expected behavior:** The images load correctly such as on the repo page and on Atom.io

**Actual behavior:** The images are not found

**Reproduces how often:** 100% of the time

### Versions

Atom    : 1.19.5
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0

apm  1.18.4
npm  3.10.10
node 6.9.5 x64
python 3.6.1
git 2.12.2

Tested on macOS Sierra 10.12.6

frasertmay:
As long as every single Atom package is from a GitHub repo, would it be okay to put an additional check in the `sanitize` function of `package-readme-view` ([here](https://github.com/atom/settings-view/blob/master/lib/package-readme-view.js#L30)) to find relative URLs and prepend the repo URL?

PR #996, opened 2017-09-14
Relative image URLs in package READMEs are now resolved correctly

frasertmay:
### Description of the Change

When rendering package READMEs, whilst sanitizing the source image, URLs are now checked to see if they are absolute or relative and if they are relative then they are resolved appropriately. For example for non-installed plugins the base URL would be the GitHub repo however if they are installed it uses the path of the installed plugin.

**Before**:
relative url before

**After**:
relative url after

### Alternate Designs

I considered attempting to match up each image in the Markdown with its counterpart on its Atom.io page or its README on GitHub and using those URLs. I chose this solution in the end as it was much simpler and in the cases I've tested still seems to work.

### Benefits

Packages using relative URLs for images in their README files as recommended by GitHub will now be rendered correctly

### Possible Drawbacks

Any URL scheme not recognized by the code I have written will be assumed to be a relative path so the image URL may be altered incorrectly.

### Applicable Issues

The applicable issue would be [here](https://github.com/atom/settings-view/issues/987)

### DIFF
--- a/lib/package-detail-view.js
+++ b/lib/package-detail-view.js
@@ -372,16 +372,22 @@ export default class PackageDetailView {
       readme = fs.readFileSync(this.readmePath, {encoding: 'utf8'})
     }
 
-    let readmeSrcBase
+    let readmeSrc
 
     if (this.pack.path) {
-      readmeSrcBase = this.pack.path
+      // If package is installed, use installed path
+      readmeSrc = this.pack.path
     } else {
+      // If package isn't installed, use url path
       let repoUrl = this.packageManager.getRepositoryUrl(this.pack)
-      readmeSrcBase = repoUrl + '/blob/master/'
+
+      // Check if URL is undefined (i.e. package is unpublished)
+      if (repoUrl) {
+        readmeSrc = repoUrl + `/blob/master/`
+      }
     }
 
-    const readmeView = new PackageReadmeView(readme, readmeSrcBase)
+    const readmeView = new PackageReadmeView(readme, readmeSrc)
     if (this.readmeView) {
       this.readmeView.element.parentElement.replaceChild(readmeView.element, this.readmeView.element)
       this.readmeView.destroy()
--- a/lib/package-detail-view.js
+++ b/lib/package-detail-view.js
@@ -372,7 +372,16 @@ export default class PackageDetailView {
       readme = fs.readFileSync(this.readmePath, {encoding: 'utf8'})
     }
 
-    const readmeView = new PackageReadmeView(readme)
+    let readmeSrcBase
+
+    if (this.pack.path) {
+      readmeSrcBase = this.pack.path
+    } else {
+      let repoUrl = this.packageManager.getRepositoryUrl(this.pack)
+      readmeSrcBase = repoUrl + '/blob/master/'
+    }
+
+    const readmeView = new PackageReadmeView(readme, readmeSrcBase)
     if (this.readmeView) {
       this.readmeView.element.parentElement.replaceChild(readmeView.element, this.readmeView.element)
       this.readmeView.destroy()
--- a/lib/package-readme-view.js
+++ b/lib/package-readme-view.js
@@ -27,7 +27,7 @@ const ATTRIBUTES_TO_REMOVE = [
   'onunload'
 ]
 
-function sanitize (html) {
+function sanitize (html, readmeSrcBase) {
   const temporaryContainer = document.createElement('div')
   temporaryContainer.innerHTML = html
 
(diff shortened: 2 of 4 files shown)


lib/package-readme-view.js
```diff
@@ -45,13 +45,42 @@ function sanitize (html) {
     checkbox.setAttribute('disabled', true)
   }
 
+  let path = require('path')
+  let urlProtocolRegex = new RegExp("^(?:[a-z]+:)?\/\/", "i")
```
50Wliu: Can't this be a simple constructed without the `RegExp` constructor, as in: `/^(?:[a-z]+:)?\/\//i`?  Then it can be inlined on L57.

lib/package-readme-view.js
```diff
@@ -45,13 +45,42 @@ function sanitize (html) {
...
+
+  for (const image of temporaryContainer.querySelectorAll('img')) {
+    let imageSrc = image.getAttribute('src')
+
+    let changeSrc = true
```
50Wliu: For extra clarity I'd rename this to `changeImageSrc`.

lib/package-readme-view.js
```diff
@@ -45,13 +45,42 @@ function sanitize (html) {
...
+    if (path.isAbsolute(imageSrc)) {
+      changeSrc = false
+    }
+
+    if (changeSrc) {
+      if (path.isAbsolute(readmeSrcBase)) {
+        // If repoUrl is a local path
+        image.setAttribute('src', path.join(readmeSrcBase, imageSrc))
+      } else {
+        // If repoUrl is a URL
```
50Wliu: Maybe expound on when this can happen in the comment

lib/package-detail-view.js
```diff
@@ -372,7 +372,16 @@ export default class PackageDetailView {
       readme = fs.readFileSync(this.readmePath, {encoding: 'utf8'})
     }
 
-    const readmeView = new PackageReadmeView(readme)
+    let readmeSrcBase
```
50Wliu: What do you think about renaming this to `readmeSrc` instead?

lib/package-detail-view.js
```diff
@@ -372,7 +372,16 @@ export default class PackageDetailView {
...
+
+    if (this.pack.path) {
+      readmeSrcBase = this.pack.path
+    } else {
+      let repoUrl = this.packageManager.getRepositoryUrl(this.pack)
+      readmeSrcBase = repoUrl + '/blob/master/'
```
50Wliu: This needs to be changed to point to the last published tag.

lib/package-detail-view.js
```diff
@@ -372,7 +372,16 @@ export default class PackageDetailView {
       readme = fs.readFileSync(this.readmePath, {encoding: 'utf8'})
     }
 
-    const readmeView = new PackageReadmeView(readme)
+    let readmeSrcBase
+
+    if (this.pack.path) {
+      readmeSrcBase = this.pack.path
+    } else {
+      let repoUrl = this.packageManager.getRepositoryUrl(this.pack)
```
50Wliu: Edge case, but this might return undefined if the repository no longer exists or the package has been unpublished.  You might want to make sure that this works in those cases.

frasertmay: Just to clarify, is it `getRepositoryUrl()` which could return undefined?

lib/package-readme-view.js
frasertmay: Like this?
```javascript
// If repoUrl is a local path (i.e. package is installed)
// If repoUrl is a URL (i.e. package isn't installed)
```

lib/package-detail-view.js
frasertmay: If you think that's better than I'm happy to change it. Would you want it changed in `package-readme-view.js` as well?

lib/package-detail-view.js
frasertmay: Do you know if there's a simple way to do that?

iolsen:
We'd love to see this change go in if you can address @50Wliu's feedback.

frasertmay:
I've addressed most of it, I'm just waiting for some clarification on a couple of points

50Wliu:
Sorry, I'll try to get back to you soon!

frasertmay:
Cool, no rush

lib/package-readme-view.js
frasertmay: Done in latest commit

lib/package-readme-view.js
frasertmay: Done in latest commit

lib/package-readme-view.js
frasertmay: Done in latest commit

lib/package-detail-view.js
frasertmay: Done in latest commit

lib/package-detail-view.js
frasertmay: In latest commit I've added some checks to make try and catch this

damieng:
I'd love to get this merged in - there seems to be some conflicting files now though - possibly because of my change from our own sanitization to dompurify.

lib/package-detail-view.js
50Wliu: I looked around a bit but it doesn't look like atom.io's API has a way of getting the latest tag.  There's also the problem that `master` isn't guaranteed to exist (and be the default branch).  Let me see if anyone else has any ideas.

nathansobo:
I'm going to attempt to resolve these conflicts. 🤞

50Wliu:
@nathansobo Just so that you're aware of my last review comment that we're not sure how to resolve:
the readme location for packages that aren't installed (i.e. you're browsing in the Install view) is currently hardcoded to `readmeSrc = repoUrl + '/blob/master/'` instead of the last-published release tag.

nathansobo:
Thanks for contributing this @frasertmay!

nathansobo:
@50Wliu Hey, sorry, I missed your comment somehow before merging this 😬 . Guess the PR didn't auto-update? Should I roll this back? Still seems like a strict improvement to have some sort of README images though it won't be 100% due to the fact they could be deleted.

50Wliu:
Yeah, it's definitely an improvement.  I don't think it needs to be reverted; I'll create a follow-up issue for it.

frasertmay:
Thank's for sorting out the merge conflicts @nathansobo!

@50Wliu, if you're happy to sort out the URL issue in a separate issue/PR I'll delete the branch.

Ben3eeE:
:tada:

50Wliu:
@frasertmay Feel free to delete the branch.

PR #1005, opened 2017-10-03
Disable update notification for disabled packages [WiP]

### LINKED ISSUE #956: Update status message should not count disabled packages

JonatanPaalsson:
### Description of the Change
The purpose of this pull request is to fix the bug where a notification is given that updates are available even though the package is disabled.

This is done by looking at all the packages in the list `updates`in the `package-updates-status-view.js`and then temporarily remove all disabled packages from the list when calculating the number of packages that have available updates. 

A event is also introduced which is triggered by enabling or disabling a package. The purpose of this is so that the user is immediately informed if the package has an available update or it let's Atom know that the user is not interested to know if updates are available for the selected package.

### Alternate Designs
An alternative is the remove all the disabled packages from the list `updates` completely and then add them when the package is enabled. This however would lead to that the packages won't show up in the `available-updates` view.

### Benefits
Now all disabled packages will not trigger the update status message that appears in the lower right corner. The status message will also be updated when enabling or disabling a package.

### Possible Drawbacks

The function `updateTile`becomes a bit more complex and also have some impact on performance depending on how many packages that are in the `updates` list.
### Applicable Issues

Fixes #956

### DIFF
--- a/lib/package-updates-status-view.js
+++ b/lib/package-updates-status-view.js
@@ -27,6 +27,7 @@ export default class PackageUpdatesStatusView {
     this.disposables.add(packageManager.on('package-updating theme-updating', ({pack, error}) => { this.onPackageUpdating(pack) }))
     this.disposables.add(packageManager.on('package-updated theme-updated package-uninstalled theme-uninstalled', ({pack, error}) => { this.onPackageUpdated(pack) }))
     this.disposables.add(packageManager.on('package-update-failed theme-update-failed', ({pack, error}) => { this.onPackageUpdateFailed(pack) }))
+    this.disposables.add(atom.config.onDidChange('core.disabledPackages', () => { this.updateTile() }))
 
     const clickHandler = () => {
       atom.commands.dispatch(atom.views.getView(atom.workspace), 'settings-view:check-for-package-updates')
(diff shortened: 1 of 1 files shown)


JonatanPaalsson:
I'm having some problems writing specs for my implementation. My problem is trying to disable packages. I want to do something similar to:

```
describe "When a package is disabled", ->
    waitsForPromise ->
      atom.packages.disablePackage('out-dated-1')

    it "update the tile", ->
      expect(atom.packages.isPackageDisabled('out-dated-1')).toBe true
```

This probably fails because the package is not loaded, however I'm not sure. In a similar way I'm having problem loading this package. If anyone could help me out or point me in some direction I would be very grateful.

lib/package-updates-status-view.js
```diff
@@ -120,6 +121,22 @@ export default class PackageUpdatesStatusView {
...
+        }
+      }
+      for (let index = 0; index < disabledPackages.length; index++) {
+        if (this.updates.includes(disabledPackages[index])){
+          const location = this.updates.indexOf(disabledPackages[index])
+          this.updates.splice(location,1)
+        }
+      }
+    }
+
```
lee-dohm: I _believe_ that these two loops can be replaced with:

```
const updatesWithoutDisabled = this.updates.filter(update => !atom.packages.isPackageDisabled(update.name))
```

JonatanPaalsson: Thank you, I refactored the code using filter instead of for-loops and no the specs seem to be working.

PR #1010, opened 2017-10-10
Fix package detail view when behind a proxy

guilhermep:
### Description of the Change

To download packages behind a proxy or a firewall that requires trusting self-signed CA certificates, the user must set apm's configuration parameters `proxy`, `https-proxy` and `strict-ssl`. However, these settings are only applied to requests sent from apm itself. When trying to view package details in this environment, users are presented with an error because these requests are sent using the [Request](https://github.com/request/request) package, thus the three parameters are ignored.

This fixes package detail view from not being able to load when behind a proxy by using the same parameters set at apm config when sending requests.

### Alternate Designs

An alternative is to create atom's own proxy settings and pass them to apm instead of being the other way around.

### Benefits

Users behind a proxy or corporate firewall will be able to view package details.

### Possible Drawbacks
Server certificates are not verified when `strict-ssl` is set to false.

### Applicable Issues

atom/atom#15662

### DIFF
--- a/lib/atom-io-client.coffee
+++ b/lib/atom-io-client.coffee
@@ -13,6 +13,13 @@ class AtomIoClient
     @expiry = 1000 * 60 * 60 * 12
     @createAvatarCache()
     @expireAvatarCache()
+    @packageManager.isStrictSsl()
+      .then (resolve) =>
+        @strictSsl = resolve
+    @packageManager.getHttpsProxy()
+      .then (resolve) =>
+        @httpsProxy = resolve
+
 
   # Public: Get an avatar image from the filesystem, fetching it first if necessary
   avatar: (login, callback) ->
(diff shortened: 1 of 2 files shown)


ungb:
Hey @guilhermep,

There's a few linting errors if you look at the Appveyor or Travis builds that need to be addressed. 

Also, I was curious if you had any suggestions on how we can best test this?

guilhermep:
Hi,

To test this you could use [Squid's SSL Bump feature.](https://wiki.squid-cache.org/Features/SslPeekAndSplice) In my case I was behind an [Endian Firewall](https://www.endian.com/community/overview/), which does exactly that.

lee-dohm:
Would you mind adding some tests to this to ensure that we don't accidentally break it in the future?

Thanks for helping out!

jstallkamp:
What a pity that this seems abandoned. Is there any chance of integration ?

PR #1011, opened 2017-10-17
Add URI handler to show package pane

BinaryMuse:
Adds a URI handler for `atom://settings-view/show-package?package=package_name` that will open the package's page in Atom.

Depends on https://github.com/atom/atom/pull/11399

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -26,7 +26,7 @@ openPanel = (settingsView, panelName, uri) ->
   settingsView.showPanel(panelName, options)
 
 module.exports =
-  handleUrl: (parsed) ->
+  handleUri: (parsed) ->
     switch parsed.pathname
       when "/show-package" then @showPackage(parsed.query.package)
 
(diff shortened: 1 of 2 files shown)


lib/main.coffee
```diff
@@ -26,6 +26,13 @@ openPanel = (settingsView, panelName, uri) ->
   settingsView.showPanel(panelName, options)
 
 module.exports =
+  handleUri: (parsed) ->
```
50Wliu: This should be `handleURI` right?

BinaryMuse: Ah, if we want to match conventions with Atom core, yeah (I'll make that change directly on master), but this case matches what's in `package.json`.

akonwi:
What are your thoughts on extending this so some extra parameters in the URI can be passed to the package if the user decides to install it. 

This would allow some interesting experiences where we can potentially pre-configure some package configs before the package is installed. Additionally, it could be helpful for some packages to understand the source of new installs.

PR #1013, opened 2017-10-28
feat(update-panel): add summary to notification

FalkF:
### Description of the Change

Add a summary of updated packages to the restart notification.
The summary contains package name, last version and new version:

atom_settings-view_notify-summary

atom_settings-view_notify-summary_single_package

### Alternate Designs

Maybe some kind of scroll list is needed for large ammounts of updates,
since the restart button might disappear at the bottom.

### Benefits

A more compact summary compared to the possibly large large list of package cards.
The option to see pre and post update version when the update process is finished (package cards do not show the past version post update) .

### DIFF
--- a/lib/updates-panel.js
+++ b/lib/updates-panel.js
@@ -173,10 +173,10 @@ export default class UpdatesPanel {
           }
           const message = `Restart Atom to complete the update of ${successfulUpdatesCount} ${pluralizedPackages}:`
           let detail = ''
-          this.packageCards.forEach((card, index, array) => {
-            detail += `${card.pack.name}@${card.pack.version} -> ${card.pack.latestVersion}`
-            index === array.length - 1 ? null : detail += `\n`
+          this.packageCards.forEach((card) => {
+            detail += `${card.pack.name}@${card.pack.version} -> ${card.pack.latestVersion}\n`
           });
+          detail.trim()
 
           const buttons = [{
             text: 'Restart',
--- a/lib/updates-panel.js
+++ b/lib/updates-panel.js
@@ -171,13 +171,18 @@ export default class UpdatesPanel {
           if (successfulUpdatesCount > 1) {
             pluralizedPackages += 's'
           }
-          const message = `Restart Atom to complete the update of ${successfulUpdatesCount} ${pluralizedPackages}.`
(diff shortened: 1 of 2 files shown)


FalkF:
If you like this change, I can implement it for single updates too, pre merge.

lib/updates-panel.js
```diff
@@ -171,13 +171,18 @@ export default class UpdatesPanel {
           if (successfulUpdatesCount > 1) {
             pluralizedPackages += 's'
           }
-          const message = `Restart Atom to complete the update of ${successfulUpdatesCount} ${pluralizedPackages}.`
+          const message = `Restart Atom to complete the update of ${successfulUpdatesCount} ${pluralizedPackages}:`
+          let detail = ''
+          this.packageCards.forEach((card, index, array) => {
+            detail += `${card.pack.name}@${card.pack.version} -> ${card.pack.latestVersion}`
+            index === array.length - 1 ? null : detail += `\n`
```
50Wliu: I think this is potentially clearer as `if (index !== array.length - 1) detail += '\n'`.  Alternatively, you could add the `\n` onto `detail` every time and then just `detail.trim()` after the forEach, which removes the need for us to pass in `index` and `array`.

maxbrunsfeld: Good suggestion. Having an expression with side effects inside of a ternary is pretty unusual.

FalkF: Glad you seem to like it 😄
Could you confirm that what you are refering to as 'side effects' means changing a variable, so ternary expressions are meant for simple things and I should have used a good old if statement? (I'm beginner 👋 )

as-cii:
Nice work on this, @FalkF! 💯 This has been released as part of settings-view@0.254.0 and will be published in the next Atom beta.

lib/updates-panel.js
50Wliu: Well in this case the majority of the time you're hitting the "null" statement, while if you used an if-statement you wouldn't have to explicitly specify "do nothing".  In that way if you ever find yourself not using part of the ternary statement I think that's a signal to use an if instead.
(Though in this scenario we were able to also remove the if through `detail.trim()`)

Issue #764, opened 2016-04-06
Settings view does not disable cleanly


Ben3eeE:
### Prerequisites

* [X] Can you reproduce the problem in [safe mode](http://flight-manual.atom.io/hacking-atom/sections/debugging/#check-if-the-problem-shows-up-in-safe-mode)?
* [X] Are you running the [latest version of Atom](http://flight-manual.atom.io/hacking-atom/sections/debugging/#update-to-the-latest-version)?
* [X] Did you check the [debugging guide](flight-manual.atom.io/hacking-atom/sections/debugging/)?
* [X] Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?
* [X] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
* [X] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Aatom+-repo%3Aatom%2Felectron) to see if your bug or enhancement is already reported?

For more information on how to write a good [bug report](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-bug-report) or [enhancement request](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-enhancement-suggestion), see the `CONTRIBUTING` guide.

### Description

Disabling settings-view package has some issues. Application:show-settings command still exists in palette and help menu and executing it opens a weird atom:\config file that causes atom to hang when saving it.

Same issue different package: https://github.com/atom/about/issues/22

### Steps to Reproduce

1. Open atom
2. Disable settings-view
3. (optional) Restart atom
4. Run application:show-settings from command palette or default keybinding ctrl+, on windows.

**Expected behavior:** Nothing or preferred atom to open the config file for editing.

**Actual behavior:** Weird non existing file opens that causes atom to hang when saving.

### Versions
1.6.X
Windows 7.

50Wliu:
Again, this is because the relevant commands as well as keybindings are defined in Atom core.
https://github.com/atom/atom/blob/5464883c1ecc581ebe698bc2722b0fe51a7ccda5/src/browser/atom-application.coffee#L207

jsejcksn:
[...]

PR #1027, opened 2017-12-11
Deactivate correctly

50Wliu:
### Requirements

### Description of the Change

Properly dispose of any disposables on deactivation to prevent lingering traces of Settings View from lying around.

### Alternate Designs

None.

### Benefits

Better package deactivation.

### Possible Drawbacks

None.

### Applicable Issues

This addresses #764 to the maximum extent possible without changing code in Atom core.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -3,6 +3,8 @@ let settingsView = null
 
 let statusView = null
 
+const {CompositeDisposable} = require('atom')
+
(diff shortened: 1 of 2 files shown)


50Wliu:
It looks like this is causing the keyboard-scrollable spec to be very flaky, so I'm reverting this until I figure out the problem.

Issue #1012, opened 2017-10-26
READMEs should fall back to latest tag, not master


50Wliu:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

#996 fixed relative image URLs in READMEs.  However, for remote READMEs we fall back to the repository URL + the master branch, when the latest published tag should be used instead.  This is because a) the readme may have undergone changes since the last version and b) `master` might not even exist.

jerthiry:
I have a solution that uses the latest tag and will create a pull request as soon as specs are written.

PR #1032, opened 2017-12-27
:bug: Use latest tag for image source URL in the readme

jerthiry:
### Description of the Change

The source for the images of the README of uninstalled packages are now taken from the latest tag instread of defaulting to repo/blob/master.

### Alternate Designs

It is also possible to fall back on the latest release but some repositories have tags and no release.

### Benefits

Some repo do not have a master branch or the README has changed since the latest tag (no more images or filenames have changed), causing the images to not appear. Here it is not the case anymore.

### Possible Drawbacks

Maybe the latest tag will be in advance compared to the available version on the package.

### Applicable Issues

READMEs should fall back to latest tag, not master #1012

### DIFF
--- a/lib/package-detail-view.js
+++ b/lib/package-detail-view.js
@@ -372,21 +372,25 @@ export default class PackageDetailView {
       readme = fs.readFileSync(this.readmePath, {encoding: 'utf8'})
     }
 
-    let readmeSrc
-
     if (this.pack.path) {
       // If package is installed, use installed path
-      readmeSrc = this.pack.path
+      let readmeSrc = this.pack.path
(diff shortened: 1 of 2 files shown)


jerthiry:
Not quite sure how to write specs for this. For now, it tests that the URL of the relative image in the README is https://github.com/example/package-with-readme/blob/master/static/image.jpg. With this fix it would be something like https://github.com/example/package-with-readme/raw/v1.0.0/static/image.jpg, but it checks the latest tag online and there is of course none for this toy example.

noname2534:
+1

PR #1034, opened 2018-01-03
Remove broken offline special-casing when retrieving cached data

### LINKED ISSUE #668: Refactor how AtomIoClient communicates that package info was not found in the cache and there's no network connectivity
This is a followup issue to https://github.com/atom/settings-view/pull/666, where @thedaniel mentioned:

> if we're going to show an error, maybe we should just pass in an error instead of an imaginary value, and things will work in a more predictable way. Of course, that requires updating the other callers that special case `{}`

### LINKED ISSUE #962: Uncaught TypeError: packages[Symbol.iterator] is not a function
[Enter steps to reproduce:]

1. Open `Settings` tab
2. Click `Install` tab

**Atom**: 1.18.0 x64
**Electron**: 1.3.15
**OS**: Mac OS X 10.12.5
**Thrown From**: [settings-view](https://github.com/atom/settings-view) package 0.249.4

### Stack Trace

Uncaught TypeError: packages[Symbol.iterator] is not a function

```
At /Applications/Atom.app/Contents/Resources/app/node_modules/settings-view/lib/install-panel.js:390

TypeError: packages[Symbol.iterator] is not a function
    at InstallPanel.addPackageViews (/Applications/Atom.app/Contents/Resources/app/node_modules/settings-view/lib/install-panel.js:390:30)
    at /Applications/Atom.app/Contents/Resources/app/node_modules/settings-view/lib/install-panel.js:461:26
    at /Applications/Atom.app/Contents/Resources/app/node_modules/settings-view/lib/atom-io-client.js:68:26
    at AtomIoClient.module.exports.AtomIoClient.fetchFromCache (/Applications/Atom.app/Contents/Resources/app/node_modules/settings-view/lib/atom-io-client.js:160:22)
    at AtomIoClient.module.exports.AtomIoClient.featuredPackages (/Applications/Atom.app/Contents/Resources/app/node_modules/settings-view/lib/atom-io-client.js:65:25)
    at InstallPanel.loadFeaturedPackages (/Applications/Atom.app/Contents/Resources/app/node_modules/settings-view/lib/install-panel.js:455:27)
    at new InstallPanel (/Applications/Atom.app/Contents/Resources/app/node_modules/settings-view/lib/install-panel.js:128:16)
    at /Applications/Atom.app/Contents/Resources/app/node_modules/settings-view/lib/settings-view.js:237:22
    at SettingsView.getOrCreatePanel (/Applications/Atom.app/Contents/Resources/app/node_modules/settings-view/lib/settings-view.js:361:25)
    at SettingsView.showPanel (/Applications/Atom.app/Contents/Resources/app/node_modules/settings-view/lib/settings-view.js:419:30)
    at HTMLDivElement.clickHandler (/Applications/Atom.app/Contents/Resources/app/node_modules/settings-view/lib/settings-view.js:186:24)
```

### Commands

```
     -1:19.9.0 editor:consolidate-selections (input.hidden-input)
     -1:19.9.0 core:cancel (input.hidden-input)
     -0:46.8.0 core:backspace (input.hidden-input)
     -0:45.9.0 fuzzy-finder:toggle-file-finder (input.hidden-input)
     -0:41 editor:delete-to-beginning-of-line (input.hidden-input)
     -0:39.2.0 editor:consolidate-selections (input.hidden-input)
     -0:39.2.0 core:cancel (input.hidden-input)
  3x -0:38.8.0 key-binding-resolver:toggle (input.hidden-input)
     -0:26.2.0 settings-view:open (input.hidden-input)
```

### Non-Core Packages

```

```

50Wliu:
### Requirements

### Description of the Change

The `fetchFromCache` method looks up if there is cached data.  Then, if it hasn't expired yet or Atom doesn't have internet access, it returns the cached data.  Otherwise, if there was no cached data and Atom wasn't online, it used to return the empty object `{}` "so callers don't crash".  Ironically, this very line causes the featured packages lookup to "crash" with an exception since it expects an array.  Therefore in this PR I have removed the special-casing and always return `null` without internet access and let the callers deal with it (they already do).

Additionally I have simplified `fetchFromCache` by making it sync (there was nothing requiring it to accept a callback) and removing an unused `options` parameter.

### Alternate Designs

I'm going to rename this "Future Aspirations".  A `getaddrinfo` is much better than throwing an exception but I would still like to see a warning rather than an error; something like "Atom is currently offline and the following results may be outdated" or "Atom is currently offline.  Try again later" if there's no results.  That would require the implementation of a new `WarningView` however and is why I decided not to do that in this PR.

### Benefits

No uncaught exceptions when loading the Install page for the first time without internet access.

### Possible Drawbacks

Rather than generating an exception, we now generate a `getaddrinfo ENOENT atom.io:443` error, but at least that's gracefully handled by settings-view.

### Applicable Issues

Fixes #962
Going to say that this also fixes #668

### DIFF
--- a/lib/atom-io-client.coffee
+++ b/lib/atom-io-client.coffee
@@ -27,27 +27,27 @@ class AtomIoClient
   # caching.
   package: (name, callback) ->
     packagePath = "packages/#{name}"
-    @fetchFromCache packagePath, (err, data) =>
-      if data
-        callback(null, data)
-      else
-        @request(packagePath, callback)
+    data = @fetchFromCache(packagePath)
+    if data
+      callback(null, data)
+    else
+      @request(packagePath, callback)
 
   featuredPackages: (callback) ->
     # TODO clean up caching copypasta
-    @fetchFromCache 'packages/featured', (err, data) =>
-      if data
-        callback(null, data)
-      else
-        @getFeatured(false, callback)
+    data = @fetchFromCache 'packages/featured'
+    if data
+      callback(null, data)
+    else
+      @getFeatured(false, callback)
 
   featuredThemes: (callback) ->
     # TODO clean up caching copypasta
-    @fetchFromCache 'themes/featured', (err, data) =>
-      if data
-        callback(null, data)
-      else
-        @getFeatured(true, callback)
+    data = @fetchFromCache 'themes/featured'
+    if data
+      callback(null, data)
+    else
+      @getFeatured(true, callback)
 
   getFeatured: (loadThemes, callback) ->
     # apm already does this, might as well use it instead of request i guess? The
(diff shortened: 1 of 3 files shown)


smashwilson:
👍 Reproduced the bug and verified the fix. Thanks!

PR #1039, opened 2018-01-08
Update URI handler args

BinaryMuse:
/cc @damieng


BinaryMuse:
Failing test is intermittent and unrelated. Merging.

PR #1040, opened 2018-01-12
[enhancement] Derive 'Report Issue' URI from bugs property

### LINKED ISSUE #966: Do not derive Report Issue link from repository url
### Prerequisites

### Description

The **Report Issue** button [derives its link from the repository url](https://github.com/atom/settings-view/blob/master/lib/package-detail-view.js#L53):

      const repoUrl = this.packageManager.getRepositoryUrl(this.pack)
      if (repoUrl) {
        shell.openExternal(`${repoUrl}/issues/new`)
      }

This effective forces developers to use GitHub as the bug tracker.

A more sensible way would be to respect the [`bugs` property in `package.json`](https://docs.npmjs.com/files/package.json#bugs).

### Steps to Reproduce

1. Open package settings of a package that defines `bugs` in its `package.json` pointing to a bug tracker other than GitHub
2. Click on *Report Issue*

**Expected behavior:** Open the bug tracker defined at `bugs.url` in a brower

**Actual behavior:** Opens `REPO/issues/new`, which doesn't exist because I don't use GitHub for issue tracking

**Reproduces how often:** always

### Versions

`atom --version` output:
```
Atom    : 1.18.0
Electron: 1.3.15
Chrome  : 52.0.2743.82
Node    : 6.5.0
```

`apm --version` output:
```
apm  1.18.1
npm  3.10.10
node 6.9.5 x64
python 2.7.6
git 2.13.0
```

FalkF:
### Description of the Change

use a new method ```getRepositoryBugUri()``` that does read from the package.json [bugs property](https://docs.npmjs.com/files/package.json#bugs).

### Benefits

closes #966 (repos without github bug tracking can now link to alternatives)

### DIFF
--- a/lib/package-manager.coffee
+++ b/lib/package-manager.coffee
@@ -432,6 +432,13 @@ class PackageManager
       repoUrl = "https://github.com/#{repoName}"
     repoUrl.replace(/\.git$/, '').replace(/\/+$/, '').replace(/^git\+/, '')
 
+  getRepositoryBugUri: ({metadata}) ->
+    {bugs} = metadata
+    bugUri = bugs ?.url ? bugs ?.email ? this.getRepositoryUrl({metadata}) + '/issues/new'
+    if bugUri.includes('@')
+      bugUri = 'mailto:' + bugUri
+    bugUri
+
   checkNativeBuildTools: ->
     new Promise (resolve, reject) =>
       apmProcess = @runCommand ['install', '--check'], (code, stdout, stderr) ->
--- a/lib/package-detail-view.js
+++ b/lib/package-detail-view.js
@@ -50,7 +50,7 @@ export default class PackageDetailView {
 
     const issueButtonClickHandler = (event) => {
       event.preventDefault()
-      var bugUri = this.packageManager.getRepositoryBugUri(this.pack)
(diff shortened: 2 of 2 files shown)


lib/package-manager.coffee
```diff
@@ -432,6 +432,13 @@ class PackageManager
       repoUrl = "https://github.com/#{repoName}"
     repoUrl.replace(/\.git$/, '').replace(/\/+$/, '').replace(/^git\+/, '')
 
+  getRepositoryBugUri: ({metadata}) ->
+    {bugs} = metadata
+    bugUri = bugs ?.url ? bugs ?.email ? this.getRepositoryUrl({metadata}) + '/issues/new'
```
FalkF: I do not really like that line 🤔

maralv: There seems to be a problem with your implementation as the [NPM docs](https://docs.npmjs.com/files/package.json#bugs) says that `bugs` may be specified as an object containing a URL and/or email, or if it's only an URL it may be specified as a string directly. What exactly does metaData containg?

maralv: I took the liberty of modifying your solution to adhere to the NPM package specification, see #1093.

FalkF:
Closing in favor of}

Issue #984, opened 2017-08-30
When searching for a package to install, exact name is excluded from the list


felzix:
_From @felzix on August 27, 2017 14:54_

### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

When searching for a package to install, using the package's exact name excludes that package from the list.

### Steps to Reproduce

1. Go to the `Install` tab in `Settings`.
2. Enter in, verbatim, `react`.

**Expected behavior:**
The `react` package is in the list of packages I could install. It should probable be the first one.

**Actual behavior:**

The `react` package isn't in the list at all. However, if I only enter in `reac` (less the `t`), the `react` package shows up.

<img width="846" alt="reac" src="https://user-images.githubusercontent.com/279660/29750902-d29f8a60-8afc-11e7-9ba7-40b5dbe61596.png">
<img width="889" alt="react" src="https://user-images.githubusercontent.com/279660/29750905-d60ef83e-8afc-11e7-9cc6-0bb2ae67aa11.png">

**Reproduces how often:** [What percentage of the time does it reproduce?]

always

### Versions
atom: 1.19.3
mac os sierra 10.12.6

_Copied from original issue: atom/atom#15464_

felzix:
_From @Ben3eeE on August 28, 2017 5:58_

Thank you for reaching out!

This issue belongs in the [settings-view](https://github.com/atom/settings-view) repository. Can you please open the issue there if one does not already exist, thank you :bowing_man:

jlvivero:
I believe this bug is not on the settings view but on apm itself.

When I run the command ./apm search "react" --json=true --packages=true > log.txt
(and after trimming the non important information) I get the following list:

[{"name":"re-re-re-react",
{"name":"react-builder",
{"name":"fancy-react",
{"name":"atom-react-mf-snippets",
{"name":"react-snippets",
{"name":"react-es6-snippets",
[...]

PR #1042, opened 2018-01-19
Display search results in order returned from apm

50Wliu:
### Requirements

### Description of the Change

This PR removes the custom sorting logic that the setting-view performs after getting search results from apm.  My reasoning is that with the [recent changes](https://github.com/atom/settings-view/issues/984#issuecomment-354875049) to apm's sorting we should no longer need to add extra sorting in settings-view.  This also means that the search results will be consistent regardless of whether the search was executed through settings-view, apm, or https://atom.io/packages/search.

### Alternate Designs

None.

### Benefits

Removal of code that had minimum benefits as well as consistency with other search methods.

### Possible Drawbacks

It is possible that some people might prefer the custom sorting settings-view did.

### Applicable Issues

None.

/cc @Ingramz do you have any opinions about this?

### DIFF
--- a/lib/atom-io-client.coffee
+++ b/lib/atom-io-client.coffee
@@ -211,7 +211,8 @@ class AtomIoClient
           reject(error)
         else
           resolve(
-            body.filter (pack) ->
-              pack.releases?.latest?.map ({readme, metadata, downloads, stargazers_count}) ->
+            body
+              .filter (pack) -> pack.releases?.latest?
+              .map ({readme, metadata, downloads, stargazers_count}) ->
                 Object.assign metadata, {readme, downloads, stargazers_count}
           )
--- a/lib/atom-io-client.coffee
+++ b/lib/atom-io-client.coffee
@@ -208,11 +208,10 @@ class AtomIoClient
         if err
           error = new Error("Searching for \u201C#{query}\u201D failed.")
           error.stderr = err.message
-          reject error
+          reject(error)
         else
           resolve(
-            body.filter (pkg) -> pkg.releases?.latest?
-                .map ({readme, metadata, downloads, stargazers_count}) ->
-                  Object.assign metadata, {readme, downloads, stargazers_count}
-                .sort (a, b) -> b.downloads - a.downloads
+            body.filter (pack) ->
+              pack.releases?.latest?.map ({readme, metadata, downloads, stargazers_count}) ->
+                Object.assign metadata, {readme, downloads, stargazers_count}
           )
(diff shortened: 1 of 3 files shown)


lib/atom-io-client.coffee
```diff
@@ -208,11 +208,10 @@ class AtomIoClient
...
-          reject error
+          reject(error)
         else
           resolve(
-            body.filter (pkg) -> pkg.releases?.latest?
-                .map ({readme, metadata, downloads, stargazers_count}) ->
-                  Object.assign metadata, {readme, downloads, stargazers_count}
-                .sort (a, b) -> b.downloads - a.downloads
+            body.filter (pack) ->
+              pack.releases?.latest?.map ({readme, metadata, downloads, stargazers_count}) ->
```
Ingramz: `.filter` and `.map` need to be on the same level: first filter is done, then map, not one inside other.

50Wliu: Ah, I must have touched that line by accident when removing `.sort`.  Then Coffeelint gave me errors and I guess I didn't notice and "fixed" it incorrectly :/.

Ingramz:
I don't mind if search worked identically everywhere (website, apm, and settings-view). I think it makes the ecosystem more consistent and predictable.

While the new ordering done serverside surfaces relevant packages without requiring assitance from the clientside (searching for [blade](https://atom.io/packages/search?q=blade) shows nicely all blade related packages on top, which I like), it feels a little off. Maybe it is because a more popular package is not a top hit, but it's not as big of a deal as all 3 results are very close to the top. 

Searching for common terms like [php](https://atom.io/packages/search?q=php) or [javascript](https://atom.io/packages/search?q=javascript) also do seem to have a clear boundary from which packages containing the search term are promoted to the top, however the ordering among them seems off. If the recent serverside improvement is essentially replicating the `addCloseMatches` and `highlightExactMatch` done in settings-view, then I would suggest sort by downloads done on the results that qualify for the `addCloseMatches` criteria.

Settings view does not implement pagination for search results, which means that it is important that good search results make it to the first page, preferably above the fold. This might become more important if all relevant search results do not fit into the first page.

In general I like where this is going and I am interested in what others think.

Ben3eeE:
We got a question about this on the forums https://discuss.atom.io/t/visibility-inconsistent-package-search/51856

daviwil:
Thanks @50Wliu!  I agree that it's better to respect atom.io's search result ordering in Atom and focus future refinements in that codebase.  Merging it!

PR #1043, opened 2018-01-24
always open existing setting item in 'settings-view:open' command

### LINKED ISSUE #723: Settings view throws an exception if opened in a different pane from where it is already open
**Steps to Reproduce:**
1. Open Atom in safe mode
1. Ensure the `settings-view` is not open
1. Split right so you have two panes open
1. In the left pane open the `settings-view` using <kbd>Ctrl-,</kbd>
1. In the right pane open the `settings-view` using <kbd>Ctrl-,</kbd>
1. **Expected:** `settings-view` to focus on the already opened `settings-view` in the left pane, **Actual:** Exception in console
image

From this state you can get weird errors like: https://github.com/atom/atom/issues/12498

Refs (Same bug in atom/about): https://github.com/atom/about/issues/30

----

Expected:

"n update(s)" and the icon disappear.  One of the Settings tabs is switched to and the "Updates" section displayed.

Actual:

"n update(s)" and the icon disappear.  Nothing else (at least visibly) happens.

Atom 1.4.0 in OS X 10.10.5.

Screen recording showing the bug:

atom-clicking-update-available-in-status-bar-does-nothing-if-settings-are-open-in-multiple-tabs

t9md:
Fix #723

### Description of the Change

Condition #723 happens is

- When workspace have two or more panes and setting-view is already opened in currently not active pane.
- User invoke `settings-view:open`(it try to open `atom://config`)

Why happens is
- `atom.worksace.open` try to open already opened `setting-view` item also in different pane

### Changes

- If setting-view is already opened in workspace, activate it's pane in opener so that `workspace.open` always open already opened item in workpace.
- This hacky help is necessary since **singleton settingView** item is opened as **different URI**(workpace.open cannot open existing seting-view item by URI). 
- It eventually simulate `searchAllPanes: true` option in `workspace.open`.

Plus(I can remove this refactoring part if requested)
- Add `SettingsView::showPanelForURI` method which open matching URI's panel

Background investigation why I chose this approach is explained in https://github.com/atom/settings-view/issues/723#issuecomment-360067800.

### Alternate Designs

Currently settingsView is singleton.
Quit singleton and instantiate settingsView always is another solution.
As a result user can open multiple settingsView on every pane like one file can open in multiple editors.
But I think this is too much/big change.

### Benefits

User no loner see exception

### Possible Drawbacks

Nothing

### Applicable Issues

#723

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -26,11 +26,11 @@ module.exports = {
   activate() {
     atom.workspace.addOpener(uri => {
       if (uri.startsWith(CONFIG_URI)) {
-        if (settingsView && !settingsView.destroyed) {
+        if (settingsView == null || settingsView.destroyed) {
+          settingsView = this.createSettingsView({uri})
+        } else {
           const pane = atom.workspace.paneForItem(settingsView)
           if (pane) pane.activate()
-        } else {
-          settingsView = this.createSettingsView({uri})
         }
 
         settingsView.showPanelForURI(uri)
--- a/lib/main.js
+++ b/lib/main.js
@@ -44,7 +44,10 @@ module.exports = {
   activate() {
     atom.workspace.addOpener(uri => {
       if (uri.startsWith(CONFIG_URI)) {
-        if (settingsView == null || settingsView.destroyed) {
+        if (settingsView && !settingsView.destroyed) {
+          const pane = atom.workspace.paneForItem(settingsView)
+          if (pane) pane.activate()
+        } else {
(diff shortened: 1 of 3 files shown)


Ben3eeE:
This might also fix https://github.com/atom/atom/issues/12498 because the repro steps include getting into a bad state by opening the settings-view in two different panes.

ungb:
Thanks @t9md for this contribution!  ✨

PR #1045, opened 2018-02-12
add 'description' for object setting

### LINKED ISSUE #857: Support description for type object/grouping
All [config types](https://atom.io/docs/api/v1.10.2/Config#supported-types) support a [title and description](https://atom.io/docs/api/v1.10.2/Config#title-and-description), [except](https://github.com/atom/settings-view/blob/b091f1df383ac8a4d6bfe8c51636b150647d323c/lib/settings-panel.coffee#L313-L326) for [type object / grouping](https://atom.io/docs/api/v1.10.2/Config#object--grouping-other-types).

It would be great if the following would show a description below the title:

``` cson
config:
  someSetting:
    type: 'object'
    description: 'Description about some setting'
    properties:
      myChildIntOption:
        type: 'integer'
        minimum: 1.5
        maximum: 11.5
```

olsonpm:
### Description of the Change

This PR adds a description to configurations of type 'object'.

Currently a 'description' for a configuration of type 'object' is ignored.  This is problematic because objects allow for a grouping or subset of options which may need a description applying to all of them.  Without the object description field, settings under a scope are forced to repeatedly describe something. 

### Alternate Designs

No alternate designs were considered.  I figured this was a straight forward change and I honestly have no clue what an alternate solution could even be.

### Benefits

- Removes the need to repeat descriptions for object properties [such as these](https://github.com/olsonpm/prettier-atom/blob/permalink-for-pr/src/config-schema.json#L180-L276)
  *Note: the above link contains a repetitive note inside each scoped config description*

### Possible Drawbacks

- Due to poor nested object UI, a description for a nested property of type 'object' will not visually be understood as applying to its scoped configurations.

### Applicable Issues

fixes #857

### DIFF
--- a/lib/settings-panel.js
+++ b/lib/settings-panel.js
@@ -590,6 +590,11 @@ function elementForObject (namespace, name, value) {
     h3.textContent = getSettingTitle(keyPath, name)
     section.appendChild(h3)
 
+    const descriptionDiv = document.createElement('div')
+    descriptionDiv.classList.add('setting-description')
+    descriptionDiv.innerHTML = getSettingDescription(keyPath)
+    section.appendChild(descriptionDiv);
+
     const div = document.createElement('div')
     div.classList.add('sub-section-body')
     for (const key of sortSettings(keyPath, value)) {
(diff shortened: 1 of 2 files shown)


noname2534:
+1

Issue #1056, opened 2018-03-31
Update text in notification says updating to 'undefined'


ungb:
### Prerequisites

* [x ] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description
Update text in notification says updating to 'undefined'

### Steps to Reproduce

1. Have a few packages that need update
2. Click on update to <version> to update the package

**Expected behavior:** Message should say '<oldversion> -> <new version>'

**Actual behavior:** says updated to undefined.

see screenshot:
![image](https://user-images.githubusercontent.com/5624870/38160691-7b75d50e-3476-11e8-850b-4d0094a08887.png)

**Reproduces how often:** [What percentage of the time does it reproduce?]
100%

### Versions
```
Atom    : 1.25.0
Electron: 1.7.11
Chrome  : 58.0.3029.110
Node    : 7.9.0
```

FalkF:
Oh, thats my code failing there :cold_sweat:
I will respond with a PR later today.

PR #1057, opened 2018-04-02
fix(restart-notification): undefined latestVersion

FalkF:
### Description of the Change

When installing a package from [git url](https://github.com/atom/apm/pull/518) this [bug](https://github.com/atom/settings-view/issues/1056) occurs on the first update.
The second update shouldn't be affected because the package gets convertet from git style to normal as part of the update.

This PR makes the code for the detail message more robust and adds sha's for git packages:

git-packages
git-packages2

closes: #1056

To reproduce the bug reset a git package:
1. `apm install thomaslindstrom/color-picker`
2. `cd ~/.atom/packages/color-picker`
3. `g stash`
4. `g reset --hard a296114f3d0deec519a41f4c62e7fc56075b7f01`
5. `g stash pop`
6. replace the sha in `package.json`

Btw. what is the point of having git packages in the first place, when we convert them after the first update?

### DIFF
--- a/lib/package-card.js
+++ b/lib/package-card.js
@@ -211,7 +211,22 @@ export default class PackageCard {
           text: 'Restart',
           onDidClick () { return atom.restartApplication() }
         }]
-        const detail = `${this.pack.version} -> ${this.pack.latestVersion}`
+
+        let oldVersion = ''
+        let newVersion = ''
+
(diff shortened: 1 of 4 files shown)


lee-dohm:
Would it be possible for you to add some tests around this functionality? At least the bits that generate the various strings for display?

50Wliu:
Hmm, Appveyor failure looks real.

Issue #825, opened 2016-08-03
Merge "Install" section with "Packages" section


adamreisnz:
I can't count anymore how many times I've wanted to install a new package, went to settings, clicked "Packages" and started to type the package name in the search box waiting and staring at the screen while nothing happened...

The two interfaces "Packages" and "Install" are so similar that it's really hard to keep the two apart. Why don't we merge them into one, called "Packages" to avoid confusion?

To me it seems intuitive that if you want to install "packages" you go and click on "packages" first, especially since that's the first item that you see when you're scanning the menu options from top to bottom.

Even adding a "install new packages" button in the "packages" section linking to the "install" section would help tremendously I think.

(Ref: https://github.com/atom/atom/issues/12316)

simurai:
Yeah, not the first time I hear about this. :grimacing: Something to think about:

- Would the "Featured Packages" also be moved to "Packages"? Or maybe they would just show up if you filter/search? Or we remove them altogether?
- Would there be a difference between filtering and searching? Maybe just typing = filtering installed packages, hitting `enter` =  search for non installed packages. But that might be hard to discover. Would need a search button or the placeholder text could read something like: `Filter packages by name, hit enter to search for new packages`.
- After hitting `enter`, would only the search results be shown or also the already installed packages if the name matches? If the later, might not be easy to prioritize. What should be shown first.

Hey, another thought. How about moving the "Update" and "Install" to an entirely different package and therefore different pane. Not sure how to call it, but it would be similar to an "app store" where you discover, install and update packages, basically everything **new**. And for everything **existing** you would open the Settings View.

adambuczynski:
> Would the "Featured Packages" also be moved to "Packages"? Or maybe they would just show up if you filter/search? Or we remove them altogether?

I would vote for removing them. I don't think this is a feature that is required to be visible in the editor. It would be more fitting to have it listed somewhere on the Atom website in my opinion. 

Personally, I just install packages that I need, and I don't think I ever installed a package that was featured just because it was featured. But that might not mean much, as I behave like that in for example the Android app store as well. I just use it to get/find the apps I need, and don't care to try/install apps that other people mark as featured for whatever reason. 

> Would there be a difference between filtering and searching?

What you suggested might be workable. The placeholder sounds like a reasonable option.

Alternatives would be a small checkbox/toggle to the right of the input field, to toggle between filtering installed packages and searching for online packages, or the added search button like you suggested.

[...]

PR #1059, opened 2018-04-24
Merge Packages/Themes installation from Install to respective panels

### LINKED ISSUE #590: No way to return to "Featured Packages" list without restarting editor
Using Atom 1.0.2 on Windows 8.1 x64.

1) Open settings
2) Switch to the "Install" section & observe page shown
3) Type "foo" in the search box and press enter
4) (After deciding the search results didn't contain anything useful) Try to return to the original page that listed "Featured Packages" so you can install one of those

Expected:
One or more of the following should reset the install settings page, so that it shows the "featured packages" list:
- Deleting "foo" from the search field and pressing enter
- Switching to another settings page (eg keybindings) and then back to the "install" page
- Closing the settings tab entirely and then reopening it
- Or some kind or breadcrumb navigation, like there is on the themes settings page, when you go deeper into a theme's individual settings

Actual:
None of the above allows for returning to "featured packages". The only solution I found was to exit Atom and reload it. Ideally all/most of the above would work in resetting the install page.

szeck87:
### Requirements

### Description of the Change

This PR makes several changes to packages and themes, the idea being to eventually remove the Install pane.  The Packages and Themes panes each now has a button group to toggle between `Installed` and `Discover`, the latter being what is currently in the Install pane.

The `Installed` view functions the same as it does today, except that I changed the search to not fire until the Enter/Return key is pressed versus when you stop typing.  Since Installed and Discover share the same search, I went with that so it doesn't do http requests when looking for new packages/themes.  I can change that back and add/remove the subscription/disposable when toggling between the two.

The `Discover` view functions the same as the Install pane does today.  It loads Featured Packages/Themes by default, and searching and hitting Enter/Return does a search.  Once you "blank" out the search bar and hit enter, it loads the Featured again, which looks to close #590.

### Alternate Designs

I looked through all of the comments in #825 and incorporated various suggestions into this.

### Benefits

A bit cleaner Settings panel, since Install would eventually be removed.  Package and Theme related code, UI and functionality is moved into their respective panes instead of a whole different pane.  More efficient use of space.

### Possible Drawbacks

There will be confusion as to how you install new packages and themes, especially if the Install pane is removed right away.  If it's not removed right away, perhaps a message can be displayed for a couple of 1.x releases in the Install pane to say it's moving to Packages and Themes.

### Applicable Issues

#590 
#825

### DIFF
--- a/lib/installed-packages-panel.js
+++ b/lib/installed-packages-panel.js
@@ -23,6 +23,8 @@ export default class InstalledPackagesPanel extends CollapsibleSectionPanel {
     etch.initialize(this)
     this.settingsView = settingsView
     this.packageManager = packageManager
+    this.client = this.packageManager.getClient()
+    this.searchType = 'installed'
     this.items = {
       dev: new List('name'),
       core: new List('name'),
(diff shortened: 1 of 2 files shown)


lee-dohm:
Thanks very much for the effort here. We're concerned with fallout of moving something that a lot of people have gotten used to. Even though we agree that the current design isn't ideal, we would want to spend a bunch of design time to come up with the correct solution here which isn't something that we have resources to attack right now.

Let me know if you have any questions. Thanks very much for your effort and passion! :tada:

PR #1060, opened 2018-05-04
Perform search in "Install Panel" automatically after user stops typing

Tzook:
### Requirements

### Description of the Change

When searching packages in the "Packages" section in settings, it searches automatically upon typing ("onDidStopChanging").
This functionality is missing in the "Install" section. Searching should be consistent and more user friendly. When the user finishes typing, it should just search for him.

Example:
example

### Alternate Designs

The previous implementation was search by pressing "enter".

### Benefits

A better UX and more consistent experience

### Possible Drawbacks

More network requests

### DIFF
--- a/lib/install-panel.js
+++ b/lib/install-panel.js
@@ -43,10 +43,11 @@ export default class InstallPanel {
         }
       })
     )
-
-    this.disposables.add(atom.commands.add(this.refs.searchEditor.element, 'core:confirm', () => {
-      this.performSearch()
-    }))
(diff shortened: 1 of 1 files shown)


lee-dohm:
Thanks very much for the contribution!

daviwil:
Just tested this out, works great!  Thanks a lot for the contribution @Tzook :sparkles:

daviwil:
Just pushed version 0.255.0 of this package and updated Atom on `master` to use it.

PR #1066, opened 2018-07-10
Show enum description as default in tooltip

sonicdoe:
### Description of the Change

The default value of a configuration setting is shown next to it in a tooltip. Configuration settings also support an [`enum` key](https://atom.io/docs/api/v1.28.1/Config#enum) with `description` properties which will be displayed instead of the raw values. This description, however, only applies to the drop-down itself but not the tooltip. For example:

Screenshot showing the tooltip with “Default: default”

Since the `value` and `description` properties may not be similar, one can’t be sure which value is actually the default one. The “Color Profile” setting above is actually a good example: What’s `default`?

This pull request changes the tooltip to show the description instead (so it matches the drop-down’s choices):

Screenshot showing the tooltip with “Default: Use color profile configured in the operating system”

### Alternate Designs

The default value could also be marked in the drop-down itself, for example by suffixing “(default)”. The proposed version was selected because it fits with the existing style of indicating default values.

### Benefits

It will be more obvious which value is the default value, especially in cases where the raw value (`value`) and the description (`description`) aren’t similar.

### Possible Drawbacks

- It will no longer be possible to view the raw value of such a configuration setting.
- Tooltips may become quite large since descriptions are usually longer than their raw values.

### DIFF
--- a/lib/settings-panel.js
+++ b/lib/settings-panel.js
@@ -283,8 +283,12 @@ export default class SettingsPanel extends CollapsibleSectionPanel {
 
   bindTooltips () {
     const disposables = Array.from(this.element.querySelectorAll('input[id], select[id], atom-text-editor[id]')).map((element) => {
+      const schema = atom.config.getSchema(element.id)
       let defaultValue = this.valueToString(this.getDefault(element.id))
       if (defaultValue != null) {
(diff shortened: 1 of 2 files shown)


lee-dohm:
@daviwil Can you do the needful here?

@sonicdoe Thanks very much for the contribution!

Issue #1030, opened 2017-12-15
Cannot view the encoeded base64 image in package READMEs


yoonhyung:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

[Description of the issue]
Previously there was a problem with the relative path, so I used the base64 encoded image in the README file. As a result, the version of Atom version 1.22.1 worked well. However, the base64 encoded image is not visible as it is updated to 1.23.0.

So I checked the source(package-readme.view.js) and found that the sanitize function added a new line that merges the packaged path with the image source.

**Expected behavior:** [What you expect to happen]

it will show the encoeded based64 image in package REAEMEs

### Versions

Atom    : 1.22.1
Electron: 1.6.15
Chrome  : 56.0.2924.87
Node    : 7.4.0

apm  1.18.8
npm  3.10.10
node 6.9.5 x64
python 2.7.13
git 2.9.3.windows.1
visual studio 2015

### Additional Information
Before merging an image source with a package relative path, you need to check whether it is base64 as a regular expression in that source.

yoonhyung:
refer to the below code

```
    // if src contains a base64 enconding string then it must be a base64 enconding
    if (/^data:image\/.*;base64/i.test(imageSrc)) {
      changeImageSrc = false;
    }
```

maralv:
The referenced code snippet is not present in /lib/package-readme-view.js, this issue seems to be resolved.

maralv:
> refer to the below code
> 
> in package-readme-view.js
> 
> ```
> function sanitize(...) {
> ...
>     // if src contains a base64 enconding string then it must be a base64 enconding
>     if (/^data:image\/.*;base64/i.test(imageSrc)) {
>       changeImageSrc = false;
>     }
> ...
> }
> ```

The referenced code snippet is not present in the current /lib/package-readme-view.js, the issue seems to be resolved.

PR #1087, opened 2018-10-18
:bug: Fix base64 encoded image viewing in package README files.

maralv:
The sanitize function merges the package path with the image source,
a check is needed for base64 encoded images, to ensure they are properly
shown in the package readme view.

### Requirements

### Description of the Change

This PR adds a regex based check for base64 encoded images in package README files, with a similar syntax as surrounding checks.

### Alternate Designs

None.

### Benefits

Prior to this change, all base64 images in README files would break, when viewed from the settings view.

### Possible Drawbacks

None.

### Applicable Issues

Fixes https://github.com/atom/settings-view/issues/1030

### DIFF
--- a/lib/package-readme-view.js
+++ b/lib/package-readme-view.js
@@ -23,6 +23,11 @@ function sanitize (html, readmeSrc) {
       changeImageSrc = false
     }
 
+    // If src contains a base64 encoded image it must be left unchanged
+    if (/^data:image\/.*;base64/i.test(imageSrc)) {
+      changeImageSrc = false;
+    }
+
     // If path is absolute on file system it must be a local file, e.g. emoji
     if (path.isAbsolute(imageSrc)) {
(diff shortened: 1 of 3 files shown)


maralv:
The failed test is in ATOM_CHANNEL=beta and unrelated to this pull request.

50Wliu:
Thanks for your submission!

We require automated tests for all pull requests that include new or changed code. We do this so that we can ensure that we don't accidentally break your shiny new code the next time we or some other contributor submits a change. If you need help writing automated tests for Atom, check out [Discuss, the official Atom and Electron message board](https://discuss.atom.io) or [join the Atom Slack team](http://atom-slack.herokuapp.com/). There are a bunch of helpful community members that should be willing to point you in the right direction.

50Wliu:
Thanks!

PR #1089, opened 2018-10-25
Add option to display config enums as radio buttons

marnen:
### Requirements

### Description of the Change

This pull request adds support for a `radio` option in the config schema, for `enum` config values. If `radio` is `true`, then the enum will be displayed as a group of radio buttons. 

*Edit by @rsese to add screenshot from https://github.com/atom/settings-view/pull/1089#issuecomment-435422204*

image

### Alternate Designs

I chose to add an option here because it seemed like the easiest way to not break existing config files, and to maintain backwards compatibility for new config files. If a config file using the `radio: true` feature is processed by an older version of Atom, the extra key will simply be ignored and the option will be displayed as a `<select>` element.

### Benefits

For small numbers of options, radio buttons often provide a better user experience than a drop-down menu would. This is especially true if the option descriptions are on the long side.

*Edit by @rsese to add extra context from https://github.com/atom/settings-view/pull/1089#issuecomment-435248703*

If there are fewer than approximately 5 options, radio buttons show them all at a glance, without needing to click. Likewise, it takes only 1 click to select a radio button, as opposed to 2 clicks for a drop-down option. There are other reasons too, but those are the most prominent.

### Possible Drawbacks

I designed this to be as free of negative impact as I could manage.

### DIFF
--- a/lib/settings-panel.js
+++ b/lib/settings-panel.js
@@ -379,7 +379,7 @@ function elementForSetting (namespace, name, value) {
 
   let schema = atom.config.getSchema(`${namespace}.${name}`)
   if (schema && schema.enum) {
-    controls.appendChild(elementForOptions(namespace, name, value))
+    controls.appendChild(elementForOptions(namespace, name, value, {radio: schema.radio}))
   } else if (schema && schema.type === 'color') {
     controls.appendChild(elementForColor(namespace, name, value))
   } else if (_.isBoolean(value) || (schema && schema.type === 'boolean')) {
@@ -406,7 +406,7 @@ function getSettingTitle (keyPath, name) {
   return title || _.uncamelcase(name).split('.').map(_.capitalize).join(' ')
 }
 
-function elementForOptions (namespace, name, value) {
+function elementForOptions (namespace, name, value, {radio = false}) {
   let keyPath = `${namespace}.${name}`
   let schema = atom.config.getSchema(keyPath)
   let options = (schema && schema.enum) ? schema.enum : []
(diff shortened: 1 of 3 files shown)


lib/settings-panel.js
```diff
@@ -322,6 +326,13 @@ export default class SettingsPanel extends CollapsibleSectionPanel {
       } else {
         return floatValue
       }
+    } else if (type === 'integer') {
```
marnen: I might be able to unify this code with current float-parsing logic.

marnen:
Looks like the only failing test on AppVeyor is one that's been failing intermittently anyway, and so I believe it's unrelated to this pull request.

rsese:
Thanks @marnen! 

Can you share a screenshot of what this look like with your pull request?

Also:

> For small numbers of options, radio buttons often provide a better user experience than a drop-down menu would. This is especially true if the option descriptions are on the long side.

Can you share an example of this that shows why radio buttons would be better for a particular option?

marnen:
@rsese Sure, I’ll share a screenshot later when I’m back at my computer. 

> Can you share an example of this that shows why radio buttons would be better for a particular option?

If there are fewer than 5 or so options, radio buttons show them all at a glance, without needing to click. Likewise, it takes only 1 click to select a radio button, as opposed to a drop-down option. There are other reasons too, but those are the most prominent.

marnen:
See also https://blog.prototypr.io/7-rules-of-using-radio-buttons-vs-drop-down-menus-fddf50d312d1 (which I just found this minute). It reinforces what I was saying, but more clearly.

marnen:
Here's the promised screenshot. 
image

rsese:
Thanks!  I added the screenshot and the extra explanation of the benefits to the issue body and  someone from the team will take a look as soon as they can.

lee-dohm:
Would you mind fixing the merge conflict so we can review this?

marnen:
Yikes! I wasn’t aware that there was one; sorry about that. Apparently the mobile interface for GitHub doesn’t show that piece of info. I’ll be happy to.

marnen:
@lee-dohm Should be resolved now.

marnen:
I see some linter failures. I’ll fix those.

marnen:
Really? The linter wants a space between the function name and the parentheses? That seems very odd to me...

marnen:
This should be ready to merge now.

marnen:
Just checking: is there anything further preventing this from being reviewed?

rsese:
@marnen - sorry it's just a matter of workload, but when one of our engineers can review you'll see comments/questions/etc.

marnen:
Sure, understood. I just wanted to make sure that there was nothing else procedural that I had to take care of.

nathansobo:
I'm assuming you adding explicit parsing for the `integer` type because before we were allowing it to pass through as a string, which breaks logic associated with choosing the correct radio button? Just going to merge based on that assumption. Seems good to parse the integer regardless.

Thanks for the contribution and sorry for the delay.

marnen:
@nathansobo After 5 months, I’m afraid I don’t remember exactly why I put the parsing in, but I do believe that it was as you say, because choosing the correct radio button wasn’t working. 

Thanks for merging this! I believe this will help make the settings panel UI a bit more pleasant.

lib/settings-panel.js
people2963: Any assistance u can offer to help me set up & download PKCS#12.pfx certificates

PR #1101, opened 2019-01-23
增加array类型可以使用普通的逗号

233cy:
增加array类型可以使用普通的逗号

### Requirements

* Filling out the template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* All new code requires tests to ensure against regressions

PR #1103, opened 2019-01-24
Comma value in array

### LINKED ISSUE #1102: Allow comma in array setting value
### Prerequisites

### Description

An array setting cannot have a comma value

### Steps to Reproduce

1. Set an array config value to `["1", ",", "2"]` in `config.cson`
2. Settings view will show '1, ,, 2'
2. Add a array item in the settings view '1, ,, 2, 3'

**Expected behavior:** 

Settings view should show '1, \,, 2'
After changing the setting `config.cson` should be `["1", ",", "2", "3"]`

**Actual behavior:** 

Settings view shows '1, ,, 2'
After changing the setting `config.cson` is `["1", "", "2", "3"]`

**Reproduces how often:**
100%

### Versions

Atom    : 1.35.0-beta0
Electron: 2.0.16
Chrome  : 61.0.3163.100
Node    : 8.9.3

apm  2.1.3
npm  6.2.0
node 8.9.3 x64
atom 1.35.0-beta0
python 2.7.14
git 2.20.0.windows.1
visual studio 2015

### Additional Information

#1103 will fix this

UziTech:
### Description of the Change

`array` config type will allow an escaped comma in the value.

image

the config value will be `["1", ",", "2"]`

### Benefits

Comma can be used in array values.

### Possible Drawbacks

none

### Applicable Issues

fixes #1102
closes #1101

### DIFF
--- a/lib/settings-panel.js
+++ b/lib/settings-panel.js
@@ -307,7 +307,7 @@ export default class SettingsPanel extends CollapsibleSectionPanel {
       if (value.length === 0) {
         return null
       }
-      return value.map((val) => val.toString().replace(/,/g, "\\,")).join(', ')
+      return value.map((val) => val.toString().replace(/,/g, '\\,')).join(', ')
     } else if (value != null) {
       return value.toString()
     } else {
--- a/lib/settings-panel.js
+++ b/lib/settings-panel.js
@@ -304,7 +304,10 @@ export default class SettingsPanel extends CollapsibleSectionPanel {
 
   valueToString (value) {
     if (Array.isArray(value)) {
-      return value.join(', ') || null
+      if (value.length === 0) {
+        return null
+      }
+      return value.map((val) => val.toString().replace(/,/g, "\\,")).join(', ')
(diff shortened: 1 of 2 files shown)


lib/settings-panel.js
```diff
@@ -324,6 +327,15 @@ export default class SettingsPanel extends CollapsibleSectionPanel {
...
       let arrayValue = (value || '').split(',')
+      arrayValue = arrayValue.reduce((values, val) => {
+        const last = values.length - 1
+        if (last >= 0 && values[last].endsWith('\\')) {
+          values[last] = values[last].replace(/\\$/, ',') + val
+        } else {
+          values.push(val)
+        }
+        return values
+      }, [])
```
rafeca: This logic is a bit convoluted, but I cannot think of a better way to do this without lookbehind regex support (I've just tried using lookbehind regexps on Atom and it seems like its Chromium version does not support them yet).

Maybe once Atom uses Electron 3 this can be refactored 😃

PR #1104, opened 2019-01-31
Fix spacing of package versions

### LINKED ISSUE #1098: No space between name and version on package card
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:

### Description
Only reproduces on Beta and Nightly (not Stable).

There is no gap between the package name and it's version on package cards.

### Steps to Reproduce

1. Open Settings -> Packages
2. See `about1.9.1`

**Expected behavior:** [What you expect to happen]

(this is with `padding: 3pt` added to the empty span in between them with dev tools)

image

**Actual behavior:** [What actually happens]

image

**Reproduces how often:** [What percentage of the time does it reproduce?]

100%

### Versions
1.35.0-beta0
Ubuntu 18.10

### Additional Information

It seems to have been introduced relatively recently.

simurai:
### Description of the Change

This adds some space before the version of packages.

#### Before:

image

#### After:

image

### Benefits

Easier to read.

### Possible Drawbacks

None

### Applicable Issues

Fixes #1098

### DIFF
--- a/lib/package-card.js
+++ b/lib/package-card.js
@@ -88,7 +88,6 @@ export default class PackageCard {
         <div className='body'>
           <h4 className='card-name'>
             <a className='package-name' ref='packageName'>{displayName}</a>
-            <span />
             <span className='package-version'>
               <span ref='versionValue' className='value'>{String(this.pack.version)}</span>
             </span>
--- a/styles/package-card.less
+++ b/styles/package-card.less
@@ -77,6 +77,7 @@
       }
     }
     .package-name {
+      margin-right: .5em;
       font-weight: bolder;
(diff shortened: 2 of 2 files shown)


Aerijo:
@simurai Do you happen to know what might have caused it? Or is it just some arcane thing? Fix looks fine in any case.

simurai:
Yeah, when converting to Standard, the "space" in the `<span>` got missing: https://github.com/atom/settings-view/pull/1091/files#diff-249b0f5a1dd5495f0127542df8037f5cL89

Probably better to use CSS instead of markup for spacing.

PR #1105, opened 2019-01-31
Sticky breadcrumbs

simurai:
### Description of the Change

This makes the breadcrumbs sticky.

breadcrumbs

### Alternate Designs

You can use the navigation on the left to go back, but some users might want to click the breadcrumb navigation instead.

### Benefits

When scrolled down a bit, it allows to click the breadcrumbs without having to scroll up again.

### Possible Drawbacks

A bit less vertical scrolling space.

### Applicable Issues

Resolves #1097

PR #1107, opened 2019-02-01
Sticky sub-sections

simurai:
### Description of the Change

This PR makes the sub-sections "sticky". Also, the collapsed and uncollapsed icons got replaced with "chevrons" and placed in front.

sticky

### Benefits

- Easier to see what belongs to which subsection.
- Easier to see if a sub-section is collapsed or not. The fold icons are hard to differentiate.
- Sub-section can be collapsed after scrolling down.

### Possible Drawbacks

None.

### Applicable Issues

Inspired by #1105

### DIFF
--- a/lib/package-detail-view.js
+++ b/lib/package-detail-view.js
@@ -252,7 +252,7 @@ export default class PackageDetailView {
       )
     }
     return (
-      <div tabIndex='0' className='package-detail panels-item'>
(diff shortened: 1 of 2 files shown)


simurai:
Here the packages page:

sticky

Aerijo:
Now all it needs to do is remember where it was when you go in and out of a package card / other settings tab.

PR #1109, opened 2019-02-14
Escape CSON correctly

Aerijo:
### Requirements

### Description of the Change

Escapes quotes and backslashes better in the keybinding-to-string function.

In addition to linked issue, this is a keybinding I found on my own Atom that needs quote escaping
```cson
'atom-text-editor[data-grammar~=\'css\']':
  'ctrl-alt-c': 'css-declaration-sorter:sort'
```
Without this PR, it wouldn't escape the value quotes.

Also tidied up the composition part, which was really hard to read IMO. Now it's a clean single line with well defined newlines and spacing.

### Alternate Designs

Considered `CSON.stringify(...)`, but 
(1) it does a lot of extra work for little benefit, 
(2) it returns with double quotes, and 
(3) ~~[it's wrong](https://github.com/groupon/cson-parser/blob/15bfd9695c79b627bd4dd7aabe1a4cc925a42494/lib/stringify.js#L80); trying that with the raw text `\\"` (escaped backslash + `"`) ironically gives `\"`, which is an escaped `"`.~~ (Actually might not be wrong, as `\\"` would already be escaped by `JSON.stringify` to `\\\"`. But I don't want to add a new dependency just for one thing) 

Could also run `JSON.stringify` and then swap quote escapes + replace single with double quotes, but I don't think it's necessary.

### Benefits

Correctly escape `'`, `"`, and `\` in strings

### Possible Drawbacks

none

### Applicable Issues

closes https://github.com/atom/settings-view/issues/1110

### DIFF
--- a/lib/keybindings-panel.js
+++ b/lib/keybindings-panel.js
@@ -153,7 +153,7 @@ export default class KeybindingsPanel {
 
       const escapeCSON = (input) => { // CSON.stringify does a lot of unncecessary work and gives us double quotes
         return input.replace(/\\.?|'/g, (match) => {
-          if (match.charAt(0) !== '\\') return "\\'" // escape single quote
+          if (match.charAt(0) === '\'') return "\\'" // escape single quote
           if (match.length === 1) return '\\\\' // last character is backslash; escape it
 
           switch (match.charAt(1)) {
--- a/lib/keybindings-panel.js
+++ b/lib/keybindings-panel.js
@@ -155,10 +155,12 @@ export default class KeybindingsPanel {
         return input.replace(/\\.?|'/g, (match) => {
           if (match.charAt(0) !== '\\') return "\\'" // escape single quote
           if (match.length === 1) return '\\\\' // last character is backslash; escape it
-          if (match.charAt(1) === '\\') {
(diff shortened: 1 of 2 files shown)


rsese:
Almost forget to ask, can you add a test for this?  Whoever ends up reviewing will ask :smile:

Aerijo:
@rsese added specs

rafeca:
Thanks for the contribution! 😍

Since the escape logic of CSON [matches the one in JSON](https://github.com/lifthrasiir/cson#brief-introduction) (with the only difference of having to escape single quotes instead of double quotes when using single quote notation), it will be easier to just use `JSON.stringify()` and change the quoting escaping.

Your `escapeCSON` function would look like:

```js
const escapeCSON = (input) => {
  return JSON.stringify(input)
    .slice(1, -1) // Remove wrapping double quotes
    .replace(/\\"/g, '"') // Unescape double quotes
    .replace(/'/g, '\\\''); // Escape single quotes
}
```

This will also help escaping other reserved characters like `\n`, etc.

What do you think?

Aerijo:
Yeah, it feels safer to leave the majority of the escaping to the builtin JSON.

@rafeca Done, and the failed spec seems completely unrelated.

PR #1115, opened 2019-02-21
Support settings-view in docks

simurai:
### Requirements

### Description of the Change

This improves the layout of the settings-view when moved to the left/right dock.

settings

### Alternate Designs

There are a [few options](https://github.com/atom/settings-view/pull/1115#issuecomment-466282855) for the navigation.

### Benefits

When moving the settings-view to a dock, it allows you to keep the editor in sight. Great when making changes to the settings and then see the result in the editor.

settings

### Possible Drawbacks

- It is a bit experimental with just some CSS overrides.
- The Keybindings need to be scrolled horizontally and fit better in the center.
- Adds extra maintenance cost.

### Applicable Issues

N/A

### DIFF
--- a/styles/docks.less
+++ b/styles/docks.less
@@ -1,0 +1,147 @@
+@import "ui-variables";
+
+// Dock overrides
+// Allows settings to be used in left/right dock
+
+atom-dock.left .settings-view,
+atom-dock.right .settings-view {
+  // display: block;
+  flex-direction: column;
+
+  &.pane-item {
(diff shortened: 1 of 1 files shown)


simurai:
Here some navigation options

## A: Open on hover

nav1

## B: Icons only on the side

nav2

## C: Text only in a grid

nav3

Ok, I think the winner is option C. 

- Always visible text.
- No distracting hover effects
- Only takes up two rows

I'll push this version.

PR #1119, opened 2019-02-27
Use server repository instead of package defined

### LINKED ISSUE #1118: Package author shown in atom package list may not match actual author on atom.io
### Prerequisites

### Description
Package author displayed based on repository in atom packages settings may not match the actual author/publisher on atom.io, can be spoofed.

### Steps to Reproduce

1. This package - https://atom.io/packages/atom-whois
View the package in atom Packages search

**Expected behavior:** Shows actual author of package based on atom.io author

**Actual behavior:** Package author is shown as being atom, presumably based on the user in the `repository` field.
image

The package repo URL was set to atom - https://github.com/giovazz89/atom-whois/pull/4

**Reproduces how often:** 💯

### Versions

macOS
```
Atom    : 1.34.0
Electron: 2.0.16
Chrome  : 61.0.3163.100
Node    : 8.9.3
apm  2.1.3
```

Arcanemagus:
### Requirements

### Description of the Change

Use the repository URL returned from the atom.io server query as the URL of the repository, instead of the one defined by the package itself in its own metadata to prevent potential for impersonation of other authors.

Before:
image

After:
image

### Alternate Designs

Potential enhancements to this include making the atom.io server reject package versions where the metadata in the package doesn't agree with the repository the version is coming from. Since this hasn't been in place from the start though we need to either purge all packages from atom.io that don't match or implement something within Atom to hide this.

### Benefits

It should now be impossible to impersonate other authors within Atom's search results when submitting a package.

### Possible Drawbacks

The repository defined in the package's metadata from `package.json` is now entirely ignored in query results. I'm not sure what potential downsides there are to this. It may pose an additional barrier to migrating a package to an another user.

The entire `search` method is currently untested, so this change may have side-effects that are unaccounted for.

### Applicable Issues

Fixes #1118.

### DIFF
--- a/lib/atom-io-client.coffee
+++ b/lib/atom-io-client.coffee
@@ -208,8 +208,8 @@ class AtomIoClient
             body = JSON.parse(body)
             resolve(
               body.filter (pkg) -> pkg.releases?.latest?
-                  .map ({readme, metadata, downloads, stargazers_count}) ->
-                    Object.assign metadata, {readme, downloads, stargazers_count}
+                  .map ({readme, metadata, downloads, stargazers_count, repository}) ->
+                    Object.assign metadata, {readme, downloads, stargazers_count, repository: repository.url}
             )
           catch e
             error = new Error("Searching for \u201C#{query}\u201D failed.")


Arcanemagus:
Hmmm, this fixes it in the search results, however the package still displays the wrong information if it is installed. Since the server side information is no longer available I don't think there is anything that can be done about that though.

Issue #973, opened 2017-08-06
Flaky test: It displays the snippets registered by the package


50Wliu:
Decided to join the flaky test party.  Seen at https://ci.appveyor.com/project/Atom/settings-view/build/304/job/djckyqpa13hv2mhj:
```
InstalledPackageView
  it displays the snippets registered by the package
    timeout: timed out after 5000 msec waiting for Snippets table children to contain 2 items
```

50Wliu:
This is _still_ somehow not fixed.  After two-three attempts I'm not too sure where I'm going wrong.
https://travis-ci.org/atom/settings-view/builds/324445099

PR #1123, opened 2019-04-08
Attempt to fix unreliable tests

nathansobo:
This PR attempts intermittent failures of tests that interact with snippets. I think these failures are caused by the following race condition:

* The tests activate the `snippets` package first.
* The snippets package loads bundled snippets, then reads all loaded packages in an async function.
* The tests then activate the `language-test` package asynchronously in a `waitsFor` block.
* Most of the time, the `waitsFor` block runs before the async function that reads loaded packages in the `snippets` package, but this isn't guaranteed.
* If the `waitsFor` block runs *after* the async code in the snippets package, snippets from the `language-test` package are never loaded and the tests fail.

I think this what is happening. Hopefully I am right and CI will pass now. 🤞

### DIFF
--- a/spec/installed-package-view-spec.coffee
+++ b/spec/installed-package-view-spec.coffee
@@ -45,15 +45,15 @@ describe "InstalledPackageView", ->
     shouldRunScopeTest = parseFloat(atom.getVersion()) >= 1.33
 
     waitsForPromise ->
+      atom.packages.activatePackage(path.join(__dirname, 'fixtures', 'language-test'))
+
+    waitsForPromise ->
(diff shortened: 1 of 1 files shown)


nathansobo:
Forced push and updated the PR body with a new hypothesis.

50Wliu:
Thank you!! I've been wrangling with this so much. Should close #973, right?

nathansobo:
Yep. In the future you should always feel free to just close a linked issue that I miss on your own. Thanks for the heads up though! Closes #973.

PR #1127, opened 2019-05-09
Pass `this` into the promise function

50Wliu:
Two levels of callbacks = two fat arrows needed. I'll try to add a spec or two for this.

Fixes atom/atom#19293


rafeca:
Thanks for the quick fix! ❤️

Issue #1143, opened 2020-02-20
hidden settings


UziTech:
## Summary

Allow certain config settings to be hidden from the settings panel.

```json
{
	"config": {
		"someSetting": {
			"type": "boolean",
			"default": false,
			"hidden": true
		}
	}
}
```

## Motivation

Packages sometimes need to save configuration that the user shouldn't change. It would be nice to keep that in atom.config but not show it in the settings panel.

## Describe alternatives you've considered

Right now we just use an automatically collapsed object to hide the settings but they can still be shown.

![image](https://user-images.githubusercontent.com/97994/74965142-4a22ec00-53da-11ea-8d18-8c9d2649f54a.png)

UziTech:
@Arcanemagus Are you able to review #1144 or do you know who can?

PR #1144, opened 2020-02-20
add hidden property

UziTech:
### Description of the Change

Allow certain config settings to be hidden from the settings panel.

```cson
{
	"config": {
		"someSetting": {
			"type": "boolean",
			"default": false,
			"hidden": true
		}
	}
}
```

### Alternate Designs

Use an automatically collapsed object to hide the settings, but they can still be shown.

image

### Benefits

packages can hide configuration settings that users shouldn't change manually.

### Possible Drawbacks

none

### Applicable Issues

#1143

### DIFF
--- a/lib/settings-panel.js
+++ b/lib/settings-panel.js
@@ -392,6 +392,10 @@ function elementForSetting (namespace, name, value) {
       return document.createDocumentFragment()
     }
   }
+  let schema = atom.config.getSchema(`${namespace}.${name}`)
+  if (schema && schema.hidden) {
+    return document.createDocumentFragment()
+  }
 
   const controlGroup = document.createElement('div')
   controlGroup.classList.add('control-group')
@@ -400,7 +404,6 @@ function elementForSetting (namespace, name, value) {
   controls.classList.add('controls')
   controlGroup.appendChild(controls)
 
-  let schema = atom.config.getSchema(`${namespace}.${name}`)
   if (schema && schema.enum) {
     controls.appendChild(elementForOptions(namespace, name, value, {radio: schema.radio}))
   } else if (schema && schema.type === 'color') {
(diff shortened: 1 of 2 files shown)


UziTech:
@atom @rafeca @nathansobo any feedback?

UziTech:
@atom @50Wliu @lee-dohm @Arcanemagus @jasonrudolph 

Did this get looked at? Is it on anyone's TODO list?

UziTech:
@atom @50Wliu @lee-dohm @Arcanemagus @jasonrudolph @rafeca @nathansobo @lkashef @smashwilson 

Has anyone looked at this? I feel like it shouldn't be too much to ask for someone to look at a PR within 3 months.

the-j0k3r:
>packages can hide configuration settings that users shouldn't change manually.

You mean via UI, cause it obviously would be able to be changed manually.

UziTech:
Yes

winstliu (approved):
This seems reasonable, but I'd question why you'd need something like this in the first place.

UziTech:
Sync-settings has a few config properties that are hidden. For example, the time of the last back. I thought about storing it in local storage instead but then ran into issues because local storage is not synced between Atom windows.

idleberg:
> Use an automatically collapsed object to hide the settings, but they can still be shown

Could you explain the motivation for this? If the property is called hidden, why is the related setting visible in the UI? Hidden is the opposite of visible, this makes no sense.

Maybe you can answer the following questions:

- are these settings just _collapsed_?
- are these settings _read-only_ in the UI and can only be edited through the Config API?
- are they _hidden_ in the UI unless some condition is met (e.g. `atom.inDevMode()`)

Don't get me wrong, I understand the need for a developer to save settings that aren't meant to be edited by the user. I've done this myself using `indexedDB` or `localStorage`. If the config schema had a way to deal with "developer settings", I'd welcome it. I'm just questioning the semantics.

UziTech:
> are these settings just collapsed?

These settings would not be visible in the settings ui.

> are these settings read-only in the UI and can only be edited through the Config API?

hidden in the ui but can be changed through the config api.

> are they hidden in the UI unless some condition is met (e.g. atom.inDevMode())

always hidden

UziTech:
> Use an automatically collapsed object to hide the settings, but they can still be shown

That is an alternate design that works currently. This PR would hide it completely.

Kaktus259 (approved):

Kaktus259:
Q

PR #1161, opened 2020-09-23
Use marked instead of roaster + update marked

### LINKED ISSUE #500: Checkboxes in package Readmes are rendered as Settings View checkboxes
screenshot 2015-05-08 08 38 26

The checkboxes can also be toggled, which is a little strange (and the rendering is even stranger):
screenshot 2015-05-08 08 39 30

Compare to the styling from markdown-preview:
screenshot 2015-05-08 08 41 19

aminya:
### Description of the Change

This:
- Uses marked instead of roaster which was deprecated years ago.
- Updates marked to the latest version

### Benefits

- The old roaster was deprecated years ago and can bring many bugs and security issues. 
- The new marked has many improvements to its engine. Follows the latest GitHub behavior and has fixed a huge number of bugs, security issues, etc.

### Possible Drawbacks
N/A

### Applicable Issues
Closes #1148 
Fixes #500

### DIFF
--- a/lib/rich-description.js
+++ b/lib/rich-description.js
@@ -4,6 +4,7 @@ const renderer = new marked.Renderer()
 renderer.code = () => ''
 renderer.blockquote = () => ''
 renderer.heading = () => ''
+renderer.br = () => '<br/>'
 renderer.html = () => ''
 renderer.image = () => ''
 renderer.list = () => ''
--- a/lib/rich-description.js
+++ b/lib/rich-description.js
@@ -8,7 +8,7 @@ renderer.html = () => ''
 renderer.image = () => ''
 renderer.list = () => ''
 
-const markdown = text => marked(text, {renderer}).replace(/<p>(.*)<\/p>/, '$1').trim()
+const markdown = text => marked(text, {renderer, breaks: true}).replace(/<p>(.*)<\/p>/, '$1').trim()
 
 module.exports = {
   getSettingDescription (keyPath) {
--- a/lib/rich-description.js
+++ b/lib/rich-description.js
@@ -5,7 +5,7 @@ renderer.code = () => ''
 renderer.blockquote = () => ''
 renderer.heading = () => ''
 renderer.html = (src) => {
-  const match = src.match(/<br\/>/)
+  const match = src.match(/<br\s*\/?>/)
   if (match) {
     return `<br/>`
   }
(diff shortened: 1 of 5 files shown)


aminya:
@UziTech You are one of the main "marked" developers. It would be nice if you review this!

aminya:
GitHub does not break lines on a single line any more based on https://github.com/markedjs/marked/pull/1620. Should we update the failing test?

aminya:
> Github does support single line breaks on comments but not on readmes. I would do the same thing here. Enable line breaks for descriptions but not on the readme content.
> 

Does [this](https://github.com/atom/settings-view/pull/1161/commits/813563e274c5e3df86327535b820be62f99569a6) mean what you are suggesting?

> I noticed that marked is being used in an asynchronous way, but since marked doesn't need to connect to the file system or database it still just runs on the main thread, which can be bad for performance and lead to ReDos attacks. We should run marked in a [worker](https://marked.js.org/using_advanced#workers) or [Task](https://flight-manual.atom.io/api/v1.51.0/Task/) wherever we want it to be off the main thread.

We need to merge https://github.com/atom/atom/pull/21139 to be able to use WebWorkers. I think Task would be expensive here as it would start a new fork for every rendering.

After we merge https://github.com/atom/atom/pull/21139, we can definitely use WebWorkers for this.

UziTech:
> Does this mean what you are suggesting?

yes

> I think Task would be expensive here

I agree we should wait for web workers

UziTech:
It looks like the tests fail because the renderer is removing html but the test expects `<br/>` to be allowed. marked fixed inline html not going through the html renderer which is why it is breaking when updating marked.

The solution would be to not allow all html or update the renderer to allow `<br/>`.

aminya:
I added `br` method. If it does not work, I think we need to change `html` method to something like this:

```js
renderer.html = (src) => {
    const match = src.match(/<\/br>/);
    if (match) {
      return `</br>
    }
	return ''
}
```

aminya:
The HTML method does not capture `br`. The br method also seems to not have any effect.

aminya:
@sadick254 @darangi This is ready to go! 😃

lib/rich-description.js
```diff
@@ -4,11 +4,17 @@ const renderer = new marked.Renderer()
 renderer.code = () => ''
 renderer.blockquote = () => ''
 renderer.heading = () => ''
-renderer.html = () => ''
+renderer.html = (src) => {
+  const match = src.match(/<br\/>/)
```
UziTech, line 8: should probably make it work with any `br` tag.

```suggestion
  const match = src.match(/<br\s*\/?>/)
```

aminya: Thanks!

PR #1175, opened 2021-02-12
feat: add explicit button to skip restarting after package updating

aminya:
### Description of the Change

This pull request adds a button to skip restarting after package updating. 

### Benefits

The cross button in the top right corner does the same thing but based on my statistics, the users don't use it often, and they think restarting in the middle of their work session is forced by Atom. This button explicitly gives them the option to skip restarting after updates. This is useful if they are in the middle of a work session and restarting Atom is considered a disturbance.

### Verification
The tests are added for the "I'll do it later" button.
The rest of the tests should pass.

### DIFF
--- a/lib/package-card.js
+++ b/lib/package-card.js
@@ -208,11 +208,6 @@ export default class PackageCard {
     const updateButtonClickHandler = (event) => {
       event.stopPropagation()
       this.update().then(() => {
-        const buttons = [{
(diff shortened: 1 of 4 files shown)


aminya:
@sadick254 Please check this out.

sadick254 (approved):
This looks good. 👍

